PR #57, opened 2016-10-19
Fixed editing first name/family name fails on the dashboard

enricovianello:
Before:
- The dashboard invokes an SCIM patch operation, but sends an empty password, and this makes the database update fail.

Now:
- A patch with an empty email triggers a validation error before reaching the database
- The dashboard sends correctly the SCIM patch requests, with only the fields that are actually changed

In addiction:
- Added more validation at backend side (groups patch e.g.)
- Added tests


enricovianello:
Added selenium tests into iam-robot-testsuite.
Here is the pull request: https://github.com/indigo-iam/iam-robot-testsuite/pull/11

PR #61, opened 2016-10-26
As an autenticated user, I can change my password

enricovianello:


enricovianello:
Dashboard updates are missing.
These are only the updates backend-side.

PR #75, opened 2016-11-30
New patch interface updates

enricovianello:


iam-login-service/src/main/java/it/infn/mw/iam/api/scim/updater/UpdaterType.java
```diff
@@ -21,7 +21,8 @@
   ACCOUNT_REMOVE_X509_CERTIFICATE("Remove user's x509 certificate"),
 
   ACCOUNT_ADD_GROUP_MEMBERSHIP("Add group membership to user"),
-  ACCOUNT_REMOVE_GROUP_MEMBERSHIP("Remove user's group membership");
```
andreaceccanti: Remove all the 's. 
Examples: 
- Remove user's group membership -> Remove user from group
- Remove user's X.509 certificate -> Remove X.509 certificate from user

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/new_updater/UpdaterType.java
```diff
@@ -1,23 +0,0 @@
-package it.infn.mw.iam.api.scim.new_updater;
```
andreaceccanti: I don't like commits that destroy history and authorship. It seems git here didn't understand that the file was moved from one directory to another. Let's try to avoid this kind of errors in the future

enricovianello:
Already merged into develop

PR #95, opened 2017-02-15
Leverage the database to lookup valid user tokens

### LINKED ISSUE #94: Replace "find access/refresh token for user" inefficient query implementation with logic that leverages database joins and indexes
MitreID implementation for the "find all valid access/refresh token for user" loads all issued tokens in memory and then filters by user id:

https://github.com/indigo-iam/OpenID-Connect-Java-Spring-Server/blob/master/openid-connect-server/src/main/java/org/mitre/oauth2/service/impl/DefaultOAuth2ProviderTokenService.java#L93-L120

This is hardly scalable.
Change the code to let the DB do the joining and filtering.

andreaceccanti:
MitreID implementation for the "find all valid access/refresh token for
user" queries loads all issued tokens in memory and then filters by user
id.

This can hardly scale, so we replace the mitre implementation with our
own to let the database do the heavy lifting.

Fixes #94

### DIFF
--- a/iam-login-service/src/main/java/it/infn/mw/iam/core/IamTokenService.java
+++ b/iam-login-service/src/main/java/it/infn/mw/iam/core/IamTokenService.java
@@ -44,4 +44,15 @@ public class IamTokenService extends DefaultOAuth2ProviderTokenService {
     results.addAll(refreshTokenRepo.findValidRefreshTokensForUser(id));
     return results;
   }
+
+  @Override
+  public void revokeAccessToken(OAuth2AccessTokenEntity accessToken) {
+    accessTokenRepo.delete(accessToken);
+  }
+
+  @Override
+  public void revokeRefreshToken(OAuth2RefreshTokenEntity refreshToken) {
+    refreshTokenRepo.delete(refreshToken);
+  }
+
 }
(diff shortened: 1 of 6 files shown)


andreaceccanti:
test this please

iam-login-service/src/main/java/it/infn/mw/iam/core/IamTokenService.java
```diff
@@ -0,0 +1,47 @@
...
+      IamOAuthRefreshTokenRepository rtRepo) {
+    this.accessTokenRepo = atRepo;
+    this.refreshTokenRepo = rtRepo;
+  }
+
+
+  @Override
+  public Set<OAuth2AccessTokenEntity> getAllAccessTokensForUser(String id) {
+
+    Set<OAuth2AccessTokenEntity> results = Sets.newHashSet();
```
marcocaberletti: In Mitre implementation, LinkedHashSet is used to preserve the resultset sorting.
Using HashSet, in the web interface tokens are listed in a different order than before.

iam-login-service/src/main/java/it/infn/mw/iam/core/IamTokenService.java
```diff
@@ -0,0 +1,47 @@
...
+    results.addAll(accessTokenRepo.findValidAccessTokensForUser(id));
+    return results;
+  }
+
+
+  @Override
+  public Set<OAuth2RefreshTokenEntity> getAllRefreshTokensForUser(String id) {
+    Set<OAuth2RefreshTokenEntity> results = Sets.newHashSet();
```
marcocaberletti: Same comment, as above.

Issue #137, opened 2017-05-19
Finer grained information for some Audit Log Events


andreaceccanti:
Some IAM events fail to give all the information needed to be informed on what's going on in the system.
Example:

```json
{"@type":"AccountUpdatedEvent","timestamp":1495210657125,"category":"ACCOUNT","principal":"admin","message":"Updated account information for user test","account":{"uuid":"80e5fb8d-b7c8-451a-89ba-346ae278a66f","name":"test"},"updaterType":"Add SAML account to user","source":"ScimUserProvisioning"}
```

This event tells that a SAML account has been added to a user but does not provide information on which account.
It would be better to have a specific event for this kind of things, like: SamlAccountAddedEvent with details on which account was added for a user.

The same thing applies to the following GroupUpdatedEvent:
```json
{"@type":"GroupUpdatedEvent","timestamp":1495211111526,"category":"GROUP","principal":"admin","message":"Updated information for group Production","group":{"uuid":"c617d586-54e6-411d-8e38-64967798fa8a","name":"Production"},"updaterType":"Add group to user","source":"ScimGroupProvisioning"}
```

PR #145, opened 2017-06-01
Finer grained information for audit log

marcocaberletti:
This PR resolves issue #137. 
Main changes:
- create new account event hierarchy;
- move event creation into updaters.


iam-login-service/src/main/java/it/infn/mw/iam/api/scim/controller/ScimMeController.java
```diff
@@ -111,6 +119,7 @@ private void executePatchOperation(IamAccount account, ScimPatchOperation<ScimUs
       }
       if (u.update()) {
         hasChanged = true;
+        u.publishUpdateEvent(this, eventPublisher);
```
marcocaberletti: Is there a suggested solution? 
My first idea is to copy the updaters that change the user properties into a temporary list and iterate over this list after the user object is saved...

Issue #88, opened 2017-01-13
Support for displaying and managing hierarchical groups 


andreaceccanti:
This issue tracks the dashboard development for introducing support  for hierarchical/nested groups in the IAM (#81).

The groups part of the dashboard is modified to:

- [ ] Show the hierarchical structure of groups
- [ ] Provide the ability to specify a parent group at group creation time
- [ ] In the group display page, specify a "create subgroup" button that allows to create a subgroup

PR #148, opened 2017-06-14
Nested group management from dashboard

marcocaberletti:
This PR resolve issue #88.
Basically, this PR introduces some fixes to allow creation, removal and visualization of nested groups from the web interface.


iam-login-service/src/main/webapp/resources/iam/js/dashboard-app/controllers/groups.controller.js
```diff
@@ -36,16 +38,81 @@ function GroupsController($scope, $rootScope, $uibModal, $state, $filter,
...
+			childs.forEach(function(member){
+				if(!gc.groupMap[member.value].visited){
+					gc.groupMap[member.value].visited = true;
+					toVisit.push(member.value);
+				}
+			});
+		}
+	}
+	
+	function getChilds(item){
```
andreaceccanti: getChilds -> getChildren

andreaceccanti:
In order to simplify nested group management I would:

- include the full group name in the name field after
- changing the name field size to VARCHAR(512)
- add an index on the group name field (not sure is there)

In the dashboard, there's not anymore a distinction between fullname and name, so you can sort the groups by name and get the correct ordering for the groups page. And also other changes to the dashboard are not needed since the hierarchy for groups is evident in the name.
Moreover, we gain the ability of having nested groups with the same name, i.e.:

atlas/test
cms/test

would be allowed.

The create group logic should include a groupname limit check that takes into account the fact that the groupname is computed based on the parent group name.

Since we use slash as a hierarchy separator, the slash character shouldn't be allowed in the group name.

iam-login-service/src/main/webapp/resources/iam/js/dashboard-app/controllers/group.controller.js
```diff
@@ -43,36 +50,111 @@ function GroupController($rootScope, $state, $filter, scimFactory, $uibModal, Mo
...
+		}else{
+			group.subgroups = [];
+		}
+	}
+	
+	function memberIsAGroup(member){
+		return (member.$ref.indexOf('scim/Groups')!=-1);
+	}
+	
+	function clickToOpen() {
```
andreaceccanti: shouldn't this be called something like openAddSubgroupDialog?

iam-login-service/src/main/webapp/resources/iam/template/dashboard/groups/newgroup.html
```diff
@@ -2,18 +2,34 @@
...
-		<div class="col-xs-12">
-			<operationresult />
+	<form name="newGroupForm">
+	    <div class="form-group">
+	        <label class="control-label">Name</label>
+	        <div ng-class="{'has-error': newGroupForm.name.$dirty && newGroupForm.name.$invalid, 'has-success': newGroupForm.name.$dirty && newGroupForm.name.$valid}">
+	        	<div>
+	        		<input id="name" name="name" class="form-control" type="text" required ng-maxlength="50" placeholder="Insert group name ..." ng-model="addGroupCtrl.group.displayName">
+	        		<span class="glyphicon form-control-feedback" ng-class="{'glyphicon-remove': newGroupForm.name.$dirty && newGroupForm.name.$invalid, 'glyphicon-ok': newGroupForm.name.$dirty && newGroupForm.name.$valid}" style="right: 15px; top: 40px"></span>
+	        		<span class="help-block" ng-show="newGroupForm.name.$dirty && newGroupForm.name.$error.maxlength"> Max 50 characters</span>
```
andreaceccanti: Max 50 characters -> Group name length must not exceed 50 characters

PR #183, opened 2017-09-26
Privacy policy URL in login page & other stuff

andreaceccanti:
- Added the ability to specify, via configuration, a privacy policy URL
that is displayed in the login page
- Added the ability to customize the sign in button text
- Cosmetic UI changes to the dashboard, including the fact that the
organisation name is now always shown in the dashboard top bar


iam-login-service/src/main/webapp/WEB-INF/views/iam/login.jsp
```diff
@@ -68,9 +72,18 @@
 
       <div>
         <input type="hidden" name="${_csrf.parameterName}" value="${_csrf.token}"> 
-        <input type="submit" class="btn btn-primary btn-block" value="Login" name="submit" class="form-control">
+        <input type="submit" class="btn btn-primary btn-block" value="${loginPageConfiguration.loginButtonText}" 
+          name="submit" class="form-control">
       </div>
```
enricovianello: robot-testsuite uses `xpath=//input[@value='Login']` to click the button and login.
so it could be better to use an id and update the testsuite

PR #219, opened 2018-03-15
Use common and generic model for ListResponse into SCIM API

enricovianello:


iam-login-service/src/main/java/it/infn/mw/iam/api/scim/model/ScimListResponse.java
```diff
@@ -1,58 +1,24 @@
...
-  public ScimListResponse(List<T> resources, long totalResults, long itemsPerPage,
-      long startIndex) {
-    this.resources = resources;
-    this.totalResults = totalResults;
-    this.itemsPerPage = itemsPerPage;
-    this.startIndex = startIndex;
-  }
-
-  public long getTotalResults() {
+  private final Set<String> schemas = new HashSet<>(Collections.singletonList(SCHEMA));
```
andreaceccanti: this should be static

Issue #223, opened 2018-03-31
Impossible to enable the token exchange grant from the dashboard 


andreaceccanti:
There's a regression in 1.2.1 that prevents the enabling of the token-exchange grant-type from the
client management mitre dashboard.
As a workaround the grant type can be set "manually" from the API by doing a put with the client representation json.

PR #226, opened 2018-04-06
Add grant types to dynamic client registration page

marcocaberletti:
This PR resolves issue #223.
Basically, now both the admin client management page and the self-service client registration page have the same grant type options.

### DIFF
--- a/iam-login-service/src/main/webapp/resources/js/dynreg.js
+++ b/iam-login-service/src/main/webapp/resources/js/dynreg.js
@@ -369,7 +369,9 @@ var DynRegEditView = Backbone.View.extend({
 		'implicit': 'implicit',
 		'client_credentials': 'client_credentials',
 		'redelegate': 'urn:ietf:params:oauth:grant_type:redelegate',
-		'refresh_token': 'refresh_token'
+		'refresh_token': 'refresh_token',
+		'device': 'urn:ietf:params:oauth:grant-type:device_code',
+		'token-exchange': 'urn:ietf:params:oauth:grant-type:token-exchange'
 	},
 
 	// maps from a form-friendly name to the real response type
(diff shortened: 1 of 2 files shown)


iam-login-service/src/main/webapp/resources/js/dynreg.js
```diff
@@ -369,7 +369,9 @@ var DynRegEditView = Backbone.View.extend({
 		'implicit': 'implicit',
 		'client_credentials': 'client_credentials',
 		'redelegate': 'urn:ietf:params:oauth:grant_type:redelegate',
-		'refresh_token': 'refresh_token'
+		'refresh_token': 'refresh_token',
+		'device': 'urn:ietf:params:oauth:grant-type:device_code',
+		'token-exchange': 'urn:ietf:params:oauth:grant-type:token-exchange'
```
andreaceccanti: Also have password grant here

iam-login-service/src/main/webapp/resources/template/dynreg.html
```diff
@@ -266,6 +266,18 @@ <h1 data-i18n="client.client-form.edit"></h1>
...
+					<div>
+                        <input id="grantTypes-device" type="checkbox" name="grantTypes"
+                            <%-($.inArray("urn:ietf:params:oauth:grant-type:device_code", client.grant_types) > -1 ? 'checked' : '')%>> 
+                        <label for="grantTypes-device" class="checkbox" data-i18n="client.client-form.device">device</label>
+                    </div>
+
+                    <div>
+                        <input id="grantTypes-token-exchange" type="checkbox" name="grantTypes"
+                            <%-($.inArray("urn:ietf:params:oauth:grant-type:token-exchange", client.grant_types) > -1 ? 'checked' : '')%>> 
+                        <label for="grantTypes-token-exchange" class="checkbox" data-i18n="client.client-form.token-exchange">token exchange</label>
```
andreaceccanti: We don't want users to be entitled to self-assign the password and token-exchange grant types, so these
controls should be readonly

PR #315, opened 2020-01-28
Issue-312

marcelovilaca:
Issue-312 changed query logic so that Resourse and Registration tokens do not appear in the token page: 
Also added unit tests to assure the results are not returned

After running the code formatting some lines were modified. Should leave like this or roll back?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,4 @@
-version: '2.0'
+version: '3.5'
 
 volumes:
   db_data:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,4 @@
-version: '3.5'
+version: '2.0'
 
 volumes:
   db_data:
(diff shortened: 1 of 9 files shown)


docker-compose.yml
```diff
@@ -1,4 +1,4 @@
-version: '3.5'
+version: '2.0'
```
andreaceccanti: ??? 
Revert this.

iam-login-service/src/test/java/it/infn/mw/iam/test/api/tokens/AccessTokenGetListTests.java
```diff
@@ -16,10 +16,15 @@
 package it.infn.mw.iam.test.api.tokens;
 
 import static it.infn.mw.iam.api.tokens.TokensControllerSupport.TOKENS_MAX_PAGE_SIZE;
+import static org.hamcrest.CoreMatchers.everyItem;
```
andreaceccanti: Are these imports used?
Do you have auto organize imports on on Eclipse?

marcelovilaca: In fact no... Removing this...

PR #335, opened 2020-06-10
Allow use of custom email templates

### LINKED ISSUE #321: All newly-registered users are sent an email to set a password, no matter which method they signed up.
When a user's registration request is approved, they will be sent an email with a link to set a password for their account. Whilst this is the expected, and required, behaviour for local IAM accounts, it would be preferred if this could be disabled for users who register with an external IdP. This way we only keep passwords for the small subset of users who have local accounts, rather than potentially all users in the system.

willfurnell:
Hello,

This PR implements functionality to somewhat resolve #321 - allowing the customisation of the emails that are sent to users. By default nothing changes, and templates are gathered from the classpath. However if the environment variables
`
IAM_NOTIFICATION_USE_CUSTOM_TEMPLATES=true
IAM_NOTIFICATION_CUSTOM_TEMPLATE_LOCATION=/templates/
`

are set, then templates will be gathered from /templates/ (for example) on disk instead - allowing the service owner to customise them to their wishes. 

/templates/ should be populated with all email templates in the classpath, and then edits can be made if desired.


iam-login-service/src/main/resources/application.yml
```diff
@@ -101,6 +101,8 @@ notification:
   adminAddress: ${IAM_NOTIFICATION_ADMIN_ADDRESS:indigo-alerts@localhost}
   admin-notification-policy: ${IAM_NOTIFICATION_ADMIN_NOTIFICATION_POLICY:notify-address}
   group-manager-notification-policy: ${IAM_NOTIFICATION_GROUP_MANAGER_NOTIFICATION_POLICY:notify-gms-and-admins}
+  useCustomTemplates: ${IAM_NOTIFICATION_USE_CUSTOM_TEMPLATES:false}
```
andreaceccanti, line 104: use snake case convention, i.e. use-custom-templates over useCustomTemplates

iam-login-service/src/main/resources/application.yml
```diff
@@ -101,6 +101,8 @@ notification:
...
+  customTemplateLocation: ${IAM_NOTIFICATION_CUSTOM_TEMPLATE_LOCATION:/templates/}
```
andreaceccanti, line 105: as above

iam-login-service/src/test/java/it/infn/mw/iam/test/notification/CustomTemplateNotificationTests.java
```diff
@@ -0,0 +1,154 @@
...
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringApplicationConfiguration(classes = {IamLoginService.class, NotificationTestConfig.class,
+    CoreControllerTestSupport.class})
+@WebAppConfiguration
+@Transactional
+@WithAnonymousUser
+@TestPropertySource(properties = {
+        "notification.disable=false",
+        "notification.useCustomTemplates=true",
+        "notification.customTemplateLocation=C://Users//fou43474//iam//iam-login-service//src//test//resources//templates-on-disk/"
```
andreaceccanti, line 66: This won't likely work outside of your machine.

willfurnell: Yes that's a good point. Unfortunately anything in @TestPropertySource has to be static, otherwise I'd get the working dir at runtime. Do you have any suggestions based on your experience with Spring Boot what we could do here instead please?

iam-login-service/src/main/resources/application.yml
willfurnell: Fixed

iam-login-service/src/main/resources/application.yml
willfurnell: Fixed

iam-login-service/src/main/java/it/infn/mw/iam/notification/TransientNotificationFactory.java
```diff
@@ -245,8 +247,22 @@ public IamEmailNotification createGroupMembershipRejectedMessage(IamGroupRequest
   protected IamEmailNotification createMessage(String template, Map<String, Object> model,
       IamNotificationType messageType, String subject, List<String> receiverAddress) {
 
+    VelocityEngine ve;
+
+    if(properties.getUseCustomTemplates()) {
```
andreaceccanti, line 252: Please use these formatting conventions for Java code:
https://github.com/italiangrid/codestyle/blob/master/eclipse-google-java-codestyle-formatter.xml

the formatting rules are for Eclipse, but you should easily get a converter if you use idea.

iam-login-service/src/main/java/it/infn/mw/iam/notification/TransientNotificationFactory.java
```diff
@@ -245,8 +247,22 @@ public IamEmailNotification createGroupMembershipRejectedMessage(IamGroupRequest
...
+      ve = new VelocityEngine();
+      LOG.info("Using custom template");
```
andreaceccanti, line 254:
```java
LOG.info("Using custom templates for email notifications from folder: {}", ...);
```

would be more informative

iam-login-service/src/main/java/it/infn/mw/iam/notification/TransientNotificationFactory.java
```diff
@@ -245,8 +247,22 @@ public IamEmailNotification createGroupMembershipRejectedMessage(IamGroupRequest
   protected IamEmailNotification createMessage(String template, Map<String, Object> model,
       IamNotificationType messageType, String subject, List<String> receiverAddress) {
 
+    VelocityEngine ve;
```
andreaceccanti, line 250: OMG, I didn't see this.

You are basically creating a new engine every time a message is sent, which is not what we want.

willfurnell: Ah okay sorry I didn't realise that would be an issue. I did try and use the existing imported engine but it didn't work unfortunately.

willfurnell:
I see that you're planning on upgrading Spring Boot in #339 - as Velocity is deprecated/removed in the lastest stable version, instead of just fixing the issues with this PR - should I convert the email templating to use to the new template engine at the same time?

andreaceccanti:
Hi Will,
that would be good.
Freemarker I think is a reasonable (and supported) velocity alternative.

iam-login-service/pom.xml
```diff
@@ -303,6 +303,11 @@
       <artifactId>velocity</artifactId>
     </dependency>
 
+    <dependency>
```
andreaceccanti, line 306: Use the spring-boot-freemarker starter, i.e.

```xml
<dependency>
            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-starter-freemarker</artifactId>
</dependency>
```

iam-login-service/src/main/java/it/infn/mw/iam/config/IamConfig.java
```diff
@@ -49,13 +49,33 @@
...
+import org.springframework.ui.freemarker.FreeMarkerConfigurationFactoryBean;
+import org.springframework.web.servlet.view.freemarker.FreeMarkerViewResolver;
 
 @Configuration
 public class IamConfig {
 
   @Value("${iam.organisation.name}")
   private String iamOrganisationName;
 
+  @Bean
```
andreaceccanti, line 61: This configuration is likely handled automatically by the spring boot freemarker starter

willfurnell:
I've now pushed what should be the latest commit, so as long as that passes tests it's good to merge from me.

Issue #351, opened 2020-09-11
IAM allows submitting multiple group membership requests for the same group


andreaceccanti:
The "submit group request" is not disabled during request processing, so multiple submissions are possible.

PR #352, opened 2020-09-15
WIP: IAM allows submitting multiple group membership requests for the same group #351

marcelovilaca:
- add a disable trigger into 'submit()' controller
 
Solves: issue-351


marcelovilaca:
made a booboo

PR #395, opened 2021-07-19
WIP: Issue 391 show group labels in account home page

rmiccoli:
To show group labels in account home page, I modified two files:

**user.groups.components.js**:
I added a _groupLabels_ function that takes as input the group id and returns the labels, through the _scimFactory_ service. Then I created and populated a _userGroupLabels_ vector with the recovered labels.

**user.groups.components.html**:
I looped over the _userGroupLabels_  vector and displayed the name, prefix (if any) and value (if any) of each label per group, by using the CSS classes. These allow to reproduce the style of the labels on the group list page.

### DIFF
--- a/iam-login-service/src/main/webapp/resources/iam/apps/dashboard-app/components/user/groups/user.groups.component.js
+++ b/iam-login-service/src/main/webapp/resources/iam/apps/dashboard-app/components/user/groups/user.groups.component.js
@@ -45,8 +45,6 @@
                 });
             });
         };
-
-        // By knowing the group id and by calling the appropriate service I can recover group labels
         
         self.groupLabels = function(groupId){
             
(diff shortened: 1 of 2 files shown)


iam-login-service/src/main/webapp/resources/iam/apps/dashboard-app/components/user/groups/user.groups.component.js
```diff
@@ -22,7 +22,17 @@
...
             self.enabled = true;
-        };
+            self.userGroupLabels = {};
+            self.labelName = labelName;
+
+            angular.forEach(self.user.groups, function(g){
+                var gl = [];
+                self.groupLabels(g.value).then(function(res){
+                    gl = res;
+                    self.userGroupLabels[g.value] = gl;
```
andreaceccanti, line 32: Here you could set this directly to:
```javascript
self.userGroupLabels[g.value] = res;
```

andreaceccanti (changes requested):
LGTM with the minor changes requested.
Also, there's some duplicated code in here (i.e. the labelName function is needed also elsewhere in the dashboard) but we will deal with a rationalization on this later.

PR #404, opened 2021-08-25
Issue 396 investigate slowness in answering /jwk requests

rmiccoli:
To solve the problem, the request to the /jwk endpoint always returns the same JSON object containing the keys as these are constant.
Next, a test verifies that the /jwk endpoint returns the set of keys that have been configured for the service. It must return a JSON type object with the content that can be seen in the keystore.jwks file.


iam-login-service/src/test/java/it/infn/mw/iam/test/oauth/JWKEndpointTests.java
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  @Test
+  public void testKeys() throws Exception {
+
+    // @formatter:off
+    mvc.perform(get(endpoint))
+    .andExpect(status().isOk())
+    .andExpect(content().contentType(MediaType.APPLICATION_JSON_VALUE))
+    .andExpect(jsonPath("$.keys[0].kty").value("RSA"))
```
andreaceccanti, line 67: Add an expectation on the expected number of keys, i.e. 1

iam-login-service/src/test/java/it/infn/mw/iam/test/oauth/JWKEndpointTests.java
```diff
@@ -0,0 +1,74 @@
...
+import it.infn.mw.iam.IamLoginService;
+import it.infn.mw.iam.core.web.jwk.IamJWKSetPublishingEndpoint;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringApplicationConfiguration(classes = IamLoginService.class)
+@WebAppConfiguration
+@Transactional
+public class JWKEndpointTests {
+
+  private String endpoint = "/" + IamJWKSetPublishingEndpoint.URL;
```
andreaceccanti, line 47: this should be static and final and called ENDPOINT

andreaceccanti (changes requested):
LGTM after the suggested changes

PR #405, opened 2021-08-30
Prevent same client token exchange for offline access scope

andreaceccanti:


andreaceccanti:
Closing in favour of refactored code base that's easier to merge in develop

PR #435, opened 2021-10-29
WIP: Spring boot migration

andreaceccanti:


iam-test-client/docker/Dockerfile
```diff
@@ -1,9 +1,9 @@
-FROM openjdk:8
+FROM openjdk:11
```
sam-glendenning, line 1: I think this has been missed since the upgrade to Java 17, I've been having compatibility issues. I suggest changing this to
```
FOM openjdk:17
```

iam-test-client/docker/Dockerfile.prod
```diff
@@ -1,10 +1,10 @@
-FROM openjdk:8
+FROM openjdk:11
```
sam-glendenning, line 1: I think this has been missed since the upgrade to Java 17, I've been having compatibility issues. I suggest changing this to
```
FOM openjdk:17
```

sam-glendenning (changes requested):

iam-test-client/docker/Dockerfile
andreaceccanti: Hi Sam, you're right, thanks for the feedback, I will fix this as done for the login-service image. Note however that  we are going to drop self-managed dockerfiles in favour of buildpacks, which will be automagically built by spring without the need of providing a Dockerfile. I.e., you can get a working image by just running `mvn spring-boot:build-image` on this branch. The image naming changes a bit during this testing phase (it has a -bp suffix) but before the final release the naming will be aligned to the standard one (i.e. iam-login-service and iam-test-client).

PR #457, opened 2022-02-10
Change Private Policy to Info and Privacy Policy

marcelovilaca:
- https://github.com/indigo-iam/iam/issues/456

 FIX: CNSD-288


rmiccoli:
For now, it is not required.

PR #483, opened 2022-06-01
Fix "JWT value" setting through the new Clients management user interface

enricovianello:


iam-login-service/src/main/webapp/resources/iam/apps/dashboard-app/components/clients/client/client.component.js
```diff
@@ -86,6 +86,14 @@
                 });
             }
 
+            // avoid hidden validation errors
+            if (self.clientVal.jwks_opt == "uri") {
+                // clear jwk if "uri" option is selected
+                self.clientVal.jwk = "";
+            } else {
+                // clear jwk_uri if "value" option is selected
```
giacomini, line 94: Are you sure the "value" option is selected, here?

enricovianello: Yes, because that `jwks_opt` is related to a radio button:
https://github.com/indigo-iam/iam/blob/fix-clients/iam-login-service/src/main/webapp/resources/iam/apps/dashboard-app/components/clients/client/clientauth/clientauth.component.html#L75-L81

PR #488, opened 2022-06-28
Changes on AARC profile

rmiccoli:


iam-login-service/src/main/resources/application.yml
```diff
@@ -167,6 +167,7 @@ iam:
    
   aarc-profile:
     urn-namespace: ${IAM_AARC_PROFILE_URN_NAMESPACE:example:iam}
```
enricovianello, line 169: use iam.example as default value

iam-login-service/src/main/resources/application.yml
```diff
@@ -167,6 +167,7 @@ iam:
...
+    urn-affiliation: ${IAM_AARC_PROFILE_URN_AFFILIATION:member}
```
enricovianello, line 170: add property:
urn-nid: ${IAM_AARC_PROFILE_URN_NID:geant}

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/aarc/AarcClaimValueHelper.java
```diff
@@ -39,12 +39,15 @@ public class AarcClaimValueHelper {
   @Value("${iam.aarc-profile.urn-namespace}")
   String urnNamespace;
```
enricovianello, line 41: Import also urn-nid property

PR #496, opened 2022-07-25
Validate image URL with a custom validator

enricovianello:


iam-login-service/src/main/java/it/infn/mw/iam/config/login/LoginButtonImage.java
```diff
@@ -32,7 +32,7 @@ enum Size {
     LARGE;
   }
 
-  @URL
+  @LoginButtonImageUrl
   private String url;
```
enricovianello, line 36: Could be renamed to "UrlOrInternal" because it could be used also for fields of other classes.

PR #499, opened 2022-08-19
Add lifecycle properties with their default values to application.yml

enricovianello:


giacomini:
Remove comments from the yml file.
Remember to update documentation about the environment variables.

PR #502, opened 2022-08-22
Add scope policy management to IAM dashboard

enricovianello:


enricovianello:
We will implement this within the new ReactJs based dashboard

PR #519, opened 2022-10-20
Fix AARC-G069

rmiccoli:


giacomini:
Do we ignore backwards compatibility? G069 says "To retain compatibility with [AARC-G002], implementations MAY additionally send the same information using the eduperson_entitlement claim."

enricovianello:
Replaced by #553

Issue #481, opened 2022-05-20
Deleting a client that owns some token causes tokens page failure


enricovianello:

enricovianello:
Fixed with commit https://github.com/indigo-iam/iam/commit/9e0e5a9701a3317b77a311072ce846e84d8edaf6.
This fix revokes all the tokens owned by a client that is requested to be removed. Its related registration token has been preserved to not change a behavior represented by this test:
https://github.com/indigo-iam/iam/blob/develop/iam-login-service/src/test/java/it/infn/mw/iam/test/oauth/client_registration/ClientRegistrationTests.java#L118-L159
It expects a 404 using the registration access token after client deletion.
If we remove also the registration access token, it will get 401.
@giacomini what do you think is the best behavior when a client has been deleted?

PR #547, opened 2023-01-09
Remove orphans from database due to issue #481

enricovianello:


iam-persistence/src/main/java/db/migration/h2/V81_2___RemoveOrphanTokens.java
```diff
@@ -0,0 +1,33 @@
+/**
+ * Copyright (c) Istituto Nazionale di Fisica Nucleare (INFN). 2016-2018
```
giacomini, line 2: Change copyright year

PR #619, opened 2023-06-19
Fix update of AT/RT lifetmes

federicaagostini:
The Access Token and Refresh Token lifetimes are now shown (to IAM Admins) according with the values set by configuration.
An Admin can also modify those values via web interface.


federicaagostini:
Replaced by #620

PR #623, opened 2023-06-22
Make MySQL columns subject_dn and issuer_dn case sensitive

rmiccoli:


iam-persistence/src/main/resources/db/migration/mysql/V94__alter_x509_table.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE iam_x509_cert MODIFY subject_dn char(255) CHARACTER SET latin1 COLLATE latin1_general_cs;
```
rmiccoli, line 1: Change data type to varchar(256)

rmiccoli:
Included in #624

PR #624, opened 2023-06-27
Allow to add certificates with the same subjectDn

rmiccoli:


giacomini:
To make things more general, we should allow any unique combination of subject and issuer. For each combination only one certificate is kept. DNs are case sensitive.

iam-login-service/src/main/java/it/infn/mw/iam/api/account_linking/DefaultAccountLinkingService.java
```diff
@@ -151,7 +151,7 @@ public void linkX509Certificate(Principal authenticatedUser,
 
     Optional<IamX509Certificate> linkedCert = userAccount.getX509Certificates()
       .stream()
-      .filter(c -> c.getSubjectDn().equals(x509Credential.getSubject()))
+      .filter(c -> c.getIssuerDn().equals(x509Credential.getIssuer()))
```
giacomini, line 154: We should keep both.

giacomini: It's probably useless to set again subject and issuer at lines 160-161. And since these certificates are kept in a Set, it's probably also safer not to touch the fields that define the order in the Set.

iam-persistence/src/main/java/it/infn/mw/iam/persistence/model/IamX509Certificate.java
```diff
@@ -102,12 +102,7 @@ public boolean equals(Object obj) {
     if (getClass() != obj.getClass())
       return false;
     IamX509Certificate other = (IamX509Certificate) obj;
-    if (subjectDn == null) {
-      if (other.subjectDn != null)
-        return false;
-    } else if (!subjectDn.equals(other.subjectDn))
-      return false;
-    return true;
+    return issuerDn.equals(other.issuerDn);
```
giacomini, line 105: Equality should consider both issuer and subject. Apart from the null checks, this should be something like `return subjectDn.equals(other.subjectDn) && issuerDn.equals(other.issuerDn);`.

giacomini: We need to change also `hashCode()`, which has to consider both issuer and subject. The implementation can rely on `Objects.hashCode()` or, even better, `Objects.hash()`, i.e. `return Objects.hash(subjectDn, issuerDn);`. See for example https://www.baeldung.com/java-objects-hash-vs-objects-hashcode.

iam-persistence/src/main/resources/db/migration/mysql/V94__alter_x509_table.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE iam_x509_cert MODIFY subject_dn char(255) CHARACTER SET latin1 COLLATE latin1_general_cs;
+ALTER TABLE iam_x509_cert MODIFY issuer_dn varchar(256) CHARACTER SET latin1 COLLATE latin1_general_cs;
```
giacomini, line 2: Both `char` or `varchar`.

giacomini (changes requested):

PR #637, opened 2023-07-27
Add new JWT profile that rename 'groups' claim with 'roles'

enricovianello:
The 'kc' profile wants to allow the integration with a Keycloak environment. Having "roles" claim instead of "groups" allows integration with the OIDC client - a conteinerised MISP deployment - which requires them to map IAM users.

This PR add this "kc" profile and:
- add access-token builder tests
- add scope-profile resolver tests

### DIFF
--- a/iam-common/pom.xml
+++ b/iam-common/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <groupId>it.infn.mw</groupId>
     <artifactId>iam-parent</artifactId>
-    <version>1.8.2</version>
+    <version>1.8.3</version>
   </parent>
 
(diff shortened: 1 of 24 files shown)


enricovianello:
@jamesacris we opened this new PR and decided to rename profile from "keycloak" to "kc" because we want to avoid "copyright" problems :) You can checkout the new branch "issue-556-reloaded" if you want to contribute but the idea is that by the end of this morning we finalise tests and create a IAM docker image with these changes that can be tested. This docker image won't contain all the other stuff related to IAM v1.8.3 because it will be easier for you to upgrade your deployment with the official v1.8.3 when it will be released.

PR #640, opened 2023-07-31
Add support for displaying specific language name in federation Metadata.

### LINKED ISSUE #635: IAM defaults to the first Display Name in federation Metadata, rather than a specific language
When looking to work with Japanese universities, we have had issues with IdPs not appearing within the IAM discovery. 

Upon investigation, this seems to result from sitatutions when there are two values for `mdui:DisplayName` within the metadata, where IAM will then default to only displaying the first listed name within the discovery. 

An example of this is Keio University:

```
<mdui:DisplayName xml:lang="ja">慶應義塾大学</mdui:DisplayName>
<mdui:DisplayName xml:lang="en">Keio University</mdui:DisplayName>
```

In IAM, there are no results for "Keio University", however "慶應義塾大学" is correctly found.

The same is true for INFN. "INFN - National Institute for Nuclear Physics" is not found, where "INFN - Istituto Nazionale di Fisica Nucleare" is:

```
<mdui:DisplayName xml:lang="it">INFN - Istituto Nazionale di Fisica Nucleare</mdui:DisplayName>
<mdui:DisplayName xml:lang="en">INFN - National Institute for Nuclear Physics</mdui:DisplayName>
```

It would be good if IAM could either:

- Search all language fields, making both "Keio University" and "慶應義塾大学", for example, correct responses
- Display all values of `DisplayName`, which may result in multiple listings for an IdP
- Filter the `DisplayName` category by `xml:lang="en"`, and show only `DisplayName` vaues for a consistent language. This could be configurable, or default to `en`

Examples:

IAMExample1

IAMExample2

Sae126V:
Resolves #635

### DIFF
--- a/iam-login-service/src/main/java/it/infn/mw/iam/authn/saml/DefaultMetadataLookupService.java
+++ b/iam-login-service/src/main/java/it/infn/mw/iam/authn/saml/DefaultMetadataLookupService.java
@@ -170,9 +170,9 @@ public class DefaultMetadataLookupService
         .filter(filterForDescriptions)
         .limit(MAX_RESULTS)
         .map(description -> {
-          if (description.getDisplayNames() != null) {
-            List<LocalizedString> displayNames = description.getDisplayNames();
-
+          List<LocalizedString> displayNames = description.getDisplayNames();
+          if (displayNames != null) {
+            
             for (LocalizedString displayName : displayNames) {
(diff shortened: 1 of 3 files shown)


federicaagostini:
Looks good to me.

enricovianello (approved):

federicaagostini (approved):

PR #642, opened 2023-08-22
Consider scope matcher based on string equality for custom scopes

rmiccoli:


iam-login-service/src/test/java/it/infn/mw/iam/test/oauth/scope/ScopeRegistryTests.java
```diff
@@ -43,46 +46,59 @@
...
-    when(client.getScope()).thenReturn(Sets.newHashSet("openid","profile", "test", "test:/whatever"));
+
+    DefaultScopeMatcherRegistry matcherRegistry =
+        new DefaultScopeMatcherRegistry(newHashSet(regexpMatcher("^test:/.*$")), scopeRepo);
+
+    when(client.getScope())
+      .thenReturn(Sets.newHashSet("openid", "profile", "test", "test:/whatever"));
     Set<ScopeMatcher> matchers = matcherRegistry.findMatchersForClient(client);
-    
+
```
enricovianello, line 101: Try also a custom scope with a parametric path

federicaagostini: Another test could be when we have a scope policy on a parametric scope, and we define the same "parametric" custom scope.

Issue #448, opened 2021-12-15
IAM VOMS AA should issue a warning for an expiring user AUP signature


andreaceccanti:

rmiccoli:
Add a warning message in the user dashboard and also give the user the opportunity to re-sign the AUP.

Panos512:
Piggybagging on this one, I believe it would be helpful to let users re-sign their AUP at will and not only when the AUP has expired

belforte:
Old VOMSAdmin dismissal is very close. Can we be reassured that AUP  is properly and handled ? Is there any way to sign AUP in IAM now ? Is a mail message sent ?

PR #646, opened 2023-09-06
Issue a warning for an expiring user AUP signature

SteDev2:
Issue #448


enricovianello:
I think we can close this old PR.
We could think to add something like a banner on the dashboard (current and also new one) just to help users to better realize that their AUP is expiring. In the meanwhile we added the email notifications mechanism.

Issue #522, opened 2022-10-25
Request for an optional "Apply for an account with eduGAIN" button


norealroots:
At the moment, the only button on the IAM home page which says "apply" will direct a user to register for a local account. It would be good to have the option to display an "Apply for an account with eduGAIN" button as well.

I am aware that the flow here would be the same as if a user clicks the "Sign in with eduGAIN" button, however we often have users missing this and applying direct for local credentials as they click the only apply button. Having the second apply option would likely help avoid incorrect account applications.

Thank you!

PR #665, opened 2023-10-30
Request for an optional "Apply for an account with eduGAIN" button

Sae126V:
Resolves https://github.com/indigo-iam/iam/issues/522

Added an anchor tag with the content saying "Apply for an account with eduGAIN".

### DIFF
--- a/iam-login-service/src/main/webapp/WEB-INF/views/iam/login.jsp
+++ b/iam-login-service/src/main/webapp/WEB-INF/views/iam/login.jsp
@@ -136,8 +136,13 @@
                    Not a member?
                 </div>
                 <a class="btn btn-success btn-block" href="/start-registration">Apply for an account</a>
-                <c:if test="${loginPageConfiguration.samlEnabled}">
-                  <a class="btn btn-success btn-block" href="/saml/login">Apply for an account with eduGAIN</a>
+                <c:if test="${
+                    loginPageConfiguration.samlEnabled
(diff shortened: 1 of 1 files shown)


Sae126V:
eduGAIN_Btn

[Sample on how it looks like]

rmiccoli:
LGTM :+1:

rmiccoli (approved):
LGTM :+1:

giacomini:
Not necessarily with this PR, but we should probably use the same terminology commonly found on web services: "sign up" or "register".

rmiccoli (changes requested):
Enable the "Apply for an account with EduGAIN" button only if SAML is enabled.
e.g. `<c:if test="${loginPageConfiguration.samlEnabled}">`

Sae126V:
> Enable the "Apply for an account with EduGAIN" button only if SAML is enabled. e.g. `<c:if test="${loginPageConfiguration.samlEnabled}">`

Make sense. The will be visible only if SAML is enabled now.

rmiccoli (approved):

enricovianello:
I think there's not only EduGAIN as a SAML provider. If you configured another SAML remote provider you'll see a non sense button with the label "Apply for an account with EduGAIN" or am I wrong? And if we add this button for each OIDC/SAML provider configured the login page could contain lots of buttons. Probably we need to add a configuration parameter for each SAML endpoint configured that enables or not its own, specific "Apply for an account with ..." button.

PR #669, opened 2023-11-09
Update login form display strategy

Sae126V:
Prevents access to the login form.

Need to prevent access to the login form when admin has decided to disable(Set to false) both `localAuthenticationVisible` and  `showLinkToLocalAuthn`.


rmiccoli (commented):
Hi, I think the behaviour is already like that, isn't it?  See https://github.com/indigo-iam/iam/blob/develop/iam-login-service/src/main/webapp/WEB-INF/views/iam/login.jsp#L76.

Sae126V:
> Hi, I think the behaviour is already like that, isn't it? See https://github.com/indigo-iam/iam/blob/develop/iam-login-service/src/main/webapp/WEB-INF/views/iam/login.jsp#L76.

Say, When Admin has set loginPageMode to Hidden (_Might Happen edge case:_ Users who know the route link will still be able to access the login form even though admin doesn't want users to enter credentials to login).  Does that make sense?

rmiccoli:
> > Hi, I think the behaviour is already like that, isn't it? See https://github.com/indigo-iam/iam/blob/develop/iam-login-service/src/main/webapp/WEB-INF/views/iam/login.jsp#L76.
> 
> Say, When Admin has set loginPageMode to Hidden (_Might Happen edge case:_ Users who know the route link will still be able to access the login form even though admin doesn't want users to enter credentials to login). Does that make sense?

Ok, clear now. Yes, it makes sense to me.

rmiccoli (changes requested):
Re-thinking about this issue, the 'HIDDEN case' is already used in production and we don't want to break it. Then, you can add another property, such as DISABLED to reproduce your behavior.

Sae126V:
No worries. I thought the DISABLED Case is same as HIDDEN. If it is make no sense. I am happy to close this PR :)

enricovianello:
> Prevents access to the login form.
> 
> Need to prevent access to the login form when admin has decided to disable(Set to false) both `localAuthenticationVisible` and `showLinkToLocalAuthn`.

My only comment is that I'd change the "title" of this PR/fix. We're not preventing access, we're changing (in a correct way) the logic that hides a form. The login endpoint still login you if you present your right credentials (through a curl e.g.). Then, I'll update the PR title in order to be more clear about this. Something like "Update login form display strategy" e.g.

enricovianello (approved):

rmiccoli (approved):

federicaagostini:
LGTM

rmiccoli:
Hi @Sae126V,

we were reviewing your PR that is fine.
Currently, by setting the following properties:
* `IAM_LOCAL_AUTHN_LOGIN_PAGE_VISIBILITY = hidden`
* `IAM_LOCAL_AUTHN_ENABLED_FOR = none`

The local authentication is still shown by adding sll=y parameter, but the functionality is disabled as shown in the attached screenshot.
Screenshot from 2024-02-08 17-50-51

Let's decide together which behavior is preferred.

PR #673, opened 2023-12-01
Fix tokens default validity in seconds

enricovianello:
We were fixing an infinite AT settings with a expiration = now. We'd prefer to set the default validity in seconds set at configuration time.
We need to be sure not braking RAT logic or ID tokens.


enricovianello:
Replaced by #677

Issue #605, opened 2023-05-16
IAM should highlight how long a client has not been used


enricovianello:
This introduces the need of track a "last-used" information for a client (last token issued, last userinfo/introspection call, etc.).

giacomini:
I would rather give an administrator the opportunity to search for such clients and then delete them

giacomini:
Now that we have a draft PR, we should define well what "used" means: some access via the client credentials is done? the client fields are changed via the dashboard or via the API? something else?

darcato:
Probably it makes sense to track when a new access token is issued for a client but not all the times such access token is used to access the API. I would also track whenever the client details are updated. What do you think?

Also, I decided to track the date of last use, not the full timestamp, as it may be faster to store a less precise time indication and we don't need ms precision. Do you think it makes sense?

giacomini:
> Probably it makes sense to track when a new access token is issued for a client but not all the times such access token is used to access the API. I would also track whenever the client details are updated. What do you think?

We can start with those cases and, if needed, extend the list in the future.

> Also, I decided to track the date of last use, not the full timestamp, as it may be faster to store a less precise time indication and we don't need ms precision. Do you think it makes sense?

I don't think it makes much sense, not without an actual profiling. I even suspect that extracting the date from the timestamp is actually more work.

PR #675, opened 2023-12-04
Add last_used property to clients

darcato:
Related to issue #605 

We want to keep track of the last time a client has been used so that the long unused clients can be easily identified and deleted by the administrators.

A ``last_used`` column is added to the ``ClientDetails`` table and the API and dashboard are updated accordingly.
This requires an update to the ``ClientDetailsEntity`` on MitreID.


darcato:
I added the ``IAM_CLIENT_LAST_USED`` table with a bidirectional one to one relationship with ``CLIENT_DETAILS``. The column ``LAST_USED`` on ``CLIENT_DETAILS`` references the ``ID`` of ``IAM_CLIENT_LAST_USED`` and initially is set to ``null``.

The first time a client is used an entry is added to ``IAM_CLIENT_LAST_USED`` and its ``ID`` is written to the ``CLIENT_DETAILS(LAST_USED)``.

Then, when a client is used again the entry in the ``IAM_CLIENT_LAST_USED`` is updated with the new date. So this update does not write to the ``CLIENT_DETAILS`` table. Also, by updating the date corresponding to an existing ``ID``, we avoid adding lines to the table each time a client is used.

Still, the read operation always perform two queries to the DB (as far as I understand).

I added the ``ClientLastUsed`` entity to MitreID, since it must be visible by ``ClientDetailsEntity``. I called the table with the ``IAM_`` prefix since it is defined in the SQL migrations in the IAM project, while the entity is defined in MitreID and thus I avoided the ``Iam`` prefix on the entity name. Should I move the the SQL table creation to MitreID?

giacomini:
> I added the `IAM_CLIENT_LAST_USED` table with a bidirectional one to one relationship with `CLIENT_DETAILS`. The column `LAST_USED` on `CLIENT_DETAILS` references the `ID` of `IAM_CLIENT_LAST_USED` and initially is set to `null`.

Is it necessary to have an entry if the value is set to null?

darcato:
The value of ``LAST_USED_ID`` on ``CLIENT_DETAILS`` is set to null when there client has been created but not used yet. Anyway with the last commits I modified the one-to-one relationship to use a shared primary key. The primary key of ``IAM_CLIENT_LAST_USED`` references the primary key of ``CLIENT_DETAILS``. This way we avoid having the ``LAST_USED_ID`` column on ``CLIENT_DETAILS``.

iam-login-service/src/main/resources/application.yml
```diff
@@ -219,6 +219,9 @@ client-registration:
     default-id-token-validity-seconds: ${DEFAULT_ID_TOKEN_VALIDITY_SECONDS:600}
     default-refresh-token-validity-seconds: ${DEFAULT_REFRESH_TOKEN_VALIDITY_SECONDS:2592000}
 
+client:
+  track-last-used: ${IAM_CLIENT_TRACK_LAST_USED:false}
```
federicaagostini, line 223: I would force new IAM deployments to track the last used client (setting this property to `true` by default), but let's see what other colleagues think about it.

giacomini: Since you ask... Why should this feature be configurable?

darcato: Tracking the `last_used` date will write more often to the DB, and may cause slowdowns in issuing tokens, especially when many different clients are used. Disabling this feature can be helpful when high throughput is required.

On small/simple IAM instances this should not be a problem, so I agree that the default could be set to ``true``. Currently I set it to ``false`` since it's the backward-compatible behavior.

giacomini: Make sense. On the other hand, the feature becomes useful exactly when there are many clients, because it helps with the cleanup.

Since we don't know the impact on performance, it may make sense to be able to enable/disable it, at least for the time being. What about keeping it as an undocumented option?

iam-login-service/src/main/webapp/resources/iam/apps/dashboard-app/components/clients/clientslist/clientslist.component.html
```diff
@@ -105,6 +106,15 @@
                                 <!-- Dynamically registered -->
                                 {{ c.dynamically_registered }}
                             </td>
+                            <td class="col-xs-1">
+                                <!-- Last Used -->
+                                <div ng-if="c.last_used">
+                                    {{c.last_used | date }}
+                                </div>
+                                <div ng-if="!c.last_used" class="text-muted">
+                                    Never
```
enricovianello, line 115: I'd prefer "n.a."

enricovianello: or n/a

iam-login-service/src/main/webapp/resources/iam/apps/dashboard-app/components/clients/clientslist/clientslist.component.html
```diff
@@ -105,6 +106,15 @@
                                 <!-- Dynamically registered -->
                                 {{ c.dynamically_registered }}
                             </td>
+                            <td class="col-xs-1">
```
enricovianello, line 109: Hide this column in case IAM_CLIENT_TRACK_LAST_USED is false

iam-login-service/src/main/resources/application.yml
darcato: I change the default to ``true``.

iam-login-service/src/main/webapp/resources/iam/apps/dashboard-app/components/clients/clientslist/clientslist.component.html
darcato: Done.

iam-persistence/src/main/resources/db/migration/mysql/V100__client_last_used.sql
rmiccoli, line 1: We must change the name of the migration because a `V100__*` already exists in the `develop` branch. 
Same for h2

enricovianello: Done

PR #686, opened 2024-01-15
Use correct email address for pwd reset

federicaagostini:
When an IAM Admin changes the email address of a user, then requests to change password without refreshing the web UI, the email is correctly sent to the new address.

### DIFF
--- a/iam-login-service/src/main/webapp/resources/iam/apps/dashboard-app/components/user/password/user.password.component.js
+++ b/iam-login-service/src/main/webapp/resources/iam/apps/dashboard-app/components/user/password/user.password.component.js
@@ -50,7 +50,7 @@
         actionButtonText: 'Send password reset e-mail',
         headerText: 'Send password reset e-mail',
         bodyText:
-            `Are you sure you want to send the password reset e-mail to ${self.user.name.formatted} at ${self.userCtrl.user.emails[0].value}?`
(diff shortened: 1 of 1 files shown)


iam-login-service/src/main/webapp/resources/iam/apps/dashboard-app/components/user/password/user.password.component.js
```diff
@@ -50,7 +50,7 @@
         actionButtonText: 'Send password reset e-mail',
         headerText: 'Send password reset e-mail',
         bodyText:
-            `Are you sure you want to send the password reset e-mail to ${self.user.name.formatted}?`
+            `Are you sure you want to send the password reset e-mail to ${self.user.name.formatted} at ${self.userCtrl.user.emails[0].value}?`
```
rmiccoli, line 53: I think also `${self.user.name.formatted}` should be `${self.userCtrl.user.name.formatted}` to display the updated name correctly (e.g. if you change name or surname and then send an email to reset the password without refreshing the page)

rmiccoli:
I notice, however, that the user whose password or name/surname has been changed no longer appears in the list of users.

federicaagostini:
> I notice, however, that the user whose password or name/surname has been changed no longer appears in the list of users.

I think it appears in alphabetic order (you can use the search bar to look for the new user name for instance) -- at least it worked like this for me.

Issue #700, opened 2024-01-24
Java SQL incorrect string value


zhangxt-ihep:
Hi,

  This problem exists at `v1.8.2p2` and `v1.8.3`.
  When using IAM as an IdP to login a SP, user can choose `Institute of High Energy Physiscs (IHEP)` SAML server to emit users identities.
  It returns error code of 500 and the error is:
  ```
Request processing failed; nested exception is org.springframework.orm.jpa.JpaSystemException: Exception [EclipseLink-4002] (Eclipse Persistence Services - 2.7.9.v20210604-2c549e2208): org.eclipse.persistence.exceptions.DatabaseException Internal Exception: java.sql.SQLException: Incorrect string value: '\xE5\x94\x90\xE5\xBF\x97...' for column 'info_val' at row 1 Error Code: 1366 Call: INSERT INTO saved_user_auth_info (owner_id, info_val, info_key) VALUES (?, ?, ?) bind => [3 parameters bound] Query: DataModifyQuery(sql="INSERT INTO saved_user_auth_info (owner_id, info_val, info_key) VALUES (?, ?, ?)"); nested exception is javax.persistence.PersistenceException: Exception [EclipseLink-4002] (Eclipse Persistence Services - 2.7.9.v20210604-2c549e2208): org.eclipse.persistence.exceptions.DatabaseException Internal Exception: java.sql.SQLException: Incorrect string value: '\xE5\x94\x90\xE5\xBF\x97...' for column 'info_val' at row 1 Error Code: 1366 Call: INSERT INTO saved_user_auth_info (owner_id, info_val, info_key) VALUES (?, ?, ?) bind => [3 parameters bound] Query: DataModifyQuery(sql="INSERT INTO saved_user_auth_info (owner_id, info_val, info_key) VALUES (?, ?, ?)")
  ```
  We guess it is caused by the Chinese character insertion issues. The value '\xE5\x94\x90\xE5\xBF\x97...' is the name of our user with Chinese characters in UTF-8 format which is emitted by IHEP IdP. 
  I suggest the IAM database container defaultly turns on the support of utf-8.

  Reference: 
  https://stackoverflow.com/questions/10957238/incorrect-string-value-when-trying-to-insert-utf-8-into-mysql-via-jdbc
  https://stackoverflow.com/questions/23670754/exception-java-sql-sqlexception-incorrect-string-value-xf0-x9f-x92-xbc-for

Thanks.

rmiccoli:
PR #701

PR #701, opened 2024-01-24
Allow chinese characters in saved_user_auth_info db table

rmiccoli:
in particular, _character set_ of the `info_val` column has been changed from `latin1` to `utf8`.

This PR solves issue #700.

The error occurs when using MySQL 5.7 since:
* MySQL 5.7 has the default charset `latin1` and default collation `latin1_swedish_ci`
* MySQL 8.0 has the default charset `utf8mb4` and default collation `utf8mb4_0900_ai_ci` (which also accepts Chinese characters)

References:
* https://dev.mysql.com/doc/refman/5.7/en/charset.html
* https://dev.mysql.com/doc/refman/8.0/en/charset.html

### DIFF
--- a/iam-persistence/src/main/resources/db/migration/mysql/V100__alter_saved_user_auth_info_table.sql
+++ b/iam-persistence/src/main/resources/db/migration/mysql/V100__alter_saved_user_auth_info_table.sql
@@ -1,1 +1,1 @@
-ALTER TABLE saved_user_auth_info MODIFY info_val varchar(256) CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_ai_ci;
+ALTER TABLE saved_user_auth_info MODIFY info_val varchar(256) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
--- a/iam-persistence/src/main/resources/db/migration/mysql/V100__alter_saved_user_auth_info_table.sql
+++ b/iam-persistence/src/main/resources/db/migration/mysql/V100__alter_saved_user_auth_info_table.sql
@@ -1,0 +1,1 @@
+ALTER TABLE saved_user_auth_info MODIFY info_val varchar(256) CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_ai_ci;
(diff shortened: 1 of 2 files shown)


enricovianello:
Could you add a URL reference for the default MySQL charsets you mentioned? Thanks :)

iam-persistence/src/main/resources/db/migration/mysql/V100__alter_saved_user_auth_info_table.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE saved_user_auth_info MODIFY info_val varchar(256) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
```
enricovianello, line 1: Has this query some consequences on already present records?

enricovianello (changes requested):
Just a couple of clarifications

rmiccoli:
> Could you add a URL reference for the default MySQL charsets you mentioned? Thanks :)

URLs added in the description! :)

iam-persistence/src/main/resources/db/migration/mysql/V100__alter_saved_user_auth_info_table.sql
rmiccoli: AFAIK, there should be no problem because all characters from `utf8mb4_unicode_ci` are present in `latin1_swedish_ci`. There might be problems with conversion if there were charset columns defined as foreign keys, but this is not the case.

I tried this query on our development instance (_iam-dev_) having that table already populated, and it succeeded.

giacomini, line 1: case-insensitive?

rmiccoli: yes, I left it case-insensitive as before. Isn't it good?

giacomini: It depends on what that field represents

enricovianello (approved):

PR #766, opened 2024-05-07
Send AUP reminders and expiration email notifications

rmiccoli:
There is a 4-hour task that checks
* how many days are left before the AUP expires by sending reminder mails
* if the AUP has expired and sends an email notification about the expiration


iam-login-service/src/main/java/it/infn/mw/iam/api/aup/model/AupDTO.java
```diff
@@ -47,6 +48,11 @@ public class AupDTO {
   @Min(value = 0L, message = "Invalid AUP: signatureValidityInDays must be >= 0")
   Long signatureValidityInDays;
 
+  @NotNull(message = "Invalid AUP: aupRemindersInDays is required")
+  @Pattern(regexp = "\\d+,\\d+,\\d+",
```
enricovianello, line 52: Spaces between commas and numbers should be accepted. Start and end of line should be added to the regexp.

enricovianello:
Replaced (because included) by #787

PR #779, opened 2024-06-03
Update form-group to be dynamic for the fields

### LINKED ISSUE #759: User registration form "Notes" field usability issue
The "Notes" field in the user registration form is both **mandatory** and **empty** by default, which means the user **has** to **invent** some text to be able to submit the form.

There are a few ways to improve the UX:

- allow the field to be removed from the form (that would be best for the LHC experiments);
- make the field optional;
- provide a default content, e.g. a neutral message like "to get access to resources for my work".

If any of these improvements is easy to implement in the short term, it would be good to have that available soon, thanks!

Sae126V:
Resolves #759 

Now, Admin can able to say whether the Notes in the field is MANDATORY, OPTIONAL or HIDDEN thru env variables.

### DIFF
--- a/iam-login-service/src/main/webapp/resources/iam/apps/registration/registration.controller.js
+++ b/iam-login-service/src/main/webapp/resources/iam/apps/registration/registration.controller.js
@@ -41,6 +41,72 @@ function RegistrationController(
 
   $scope.config = undefined;
 
+  // Form `fields` with necessary properties and validations.
+  $scope.fields = {
+    name: {
+      name: "name",
+      label: "Given Name",
+      ngModelName: "givenname",
+      articleToUse: "a",
+      placeholder: "Enter your first name",
+      type: "text",
+      minlength: 2,
+      required: true,
+      validators: "",
+      showField: true,
+    },
+    surname: {
+      name: "surname",
+      label: "Family Name",
+      ngModelName: "familyname",
+      articleToUse: "a",
+      placeholder: "Enter your family name",
+      type: "text",
+      minlength: 2,
+      required: true,
+      validators: "",
+      showField: true,
+    },
+    email: {
+      name: "email",
+      label: "Email",
+      ngModelName: "email",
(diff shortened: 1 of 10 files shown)


iam-login-service/src/main/webapp/resources/iam/apps/registration/registration.controller.js
```diff
@@ -198,9 +266,30 @@ function RegistrationController(
 
   function fieldReadonly(name) {
     if ($scope.config) {
-      var field = $scope.config.fields[name];
+      var field = $scope.config.fields[name] || false;
```
garaimanoj, line 269: May be we can use `if` instead of `||`

Sae126V: Make sense to me, Manoj. Done

iam-login-service/src/main/webapp/resources/iam/apps/registration/registration.html
```diff
@@ -59,81 +59,61 @@ <h2 class="text-center">
...
+                            ng-minlength="field.minlength" ng-readonly="rc.fieldReadonly(field.name)"
+                            ng-model-options="{ debounce: field.debounceTime || 0 }" ng-attr-validator="{{field.validators}}" />
+                            
+                        <textarea
+                            ng-if="field.type === 'textarea'" class="form-control" id="{{field.name}}"
+                            name="{{field.name}}" rows="{{field.rows}}" placeholder="{{field.placeholder}}"
+                            ng-model="request[field.ngModelName]" ng-required="field.required"
+                            ng-minlength="{{field.minlength}}" ng-readonly="rc.fieldReadonly(field.name)"></textarea> 
+                        <!-- Validation messages -->
+                        <span
```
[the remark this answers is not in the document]
Sae126V: Yep. Made it more dynamic. Done

PR #781, opened 2024-06-05
Store user affiliation into user info table

### LINKED ISSUE #480: Add affiliation to user profile attributes
In several projects using IAM in France, the request was made to be able to track user affiliation as part of his user profile. It is not currently possible and no real workaround exists as the `Note` field that could be used temporarily is no longer exposed once the user has been validated. Affiliation is an important information to determine if a user is legitimate to request an IAM account, when you don't rely on an external source for identity vetting (like CERN HR database for CERN experiments).

The main requirements are:
- Add an affiliation field (free form text, granularity will be determined by the project) in the account creation form that can be made mandatory (probably some project will not want it).
- Expose this information in user profile so that it is easy to access it if needed, for security traceability for example
- Expose this information through the SCIM REST API related to users
- Prevent a user changing it without going through a validation workflow as it is a sensitive information for establishing trust. It would be acceptable to have this information displayed as a readonly attribute that only an IAM administrator can change in a first stage.

garaimanoj:
User registration page 
image

User details view for Admin
image

User details edit page for Admin
image

Response of SCIM REST api call
image


iam-persistence/src/main/java/it/infn/mw/iam/persistence/model/IamUserInfo.java
```diff
@@ -91,6 +91,9 @@ public class IamUserInfo implements Serializable {
 
   private String birthdate;
 
+  @Column(length = 4096)
```
enricovianello: we're expecting this value as a University or Institute name so having 4096 characters for this value is, in my opinion, too much. I think 200 should be greater enough to fit what we need. What do you think?

enricovianello (changes requested):
I think we can assign less than 4096 characters to this column. e.g. 200 should be enough.

iam-persistence/src/main/java/it/infn/mw/iam/persistence/model/IamUserInfo.java
garaimanoj: Updated to 200

enricovianello (approved):

enricovianello:
We must include a notice in the release notes that says: "If Affiliation is not included into your privacy notice, then it must be updated after/during IAM update"

enricovianello (approved):

PR #782, opened 2024-06-06
Add support to notes being customised

Sae126V:


Sae126V:
Hi @enricovianello @federicaagostini @rmiccoli ,

Can we have a quick chat about this PR please?

I see no matter how many test cases I have added. I am having trouble with this. I see that I am NO longer able to commit to the repo as well.


Thanks

rmiccoli:
Perhaps we can put an if statement in the email template (adminHandleRequest.ftl) so that notes are only shown if the content is not empty.
e.g.
```
<#if notes?has_content>
Notes:
${notes}
</#if>
```

Sae126V:
> Perhaps we can put an if statement in the email template (adminHandleRequest.ftl) so that notes are only shown if the content is not empty. e.g.
> 
> ```
> <#if notes?has_content>
> Notes:
> ${notes}
> </#if>
> ```

Ahh! You are right @rmiccoli 👍 . Added in 14800ca5734c3d54956ef00e1f152c0c07bbe600

iam-login-service/src/main/resources/application.yml
```diff
@@ -154,6 +154,9 @@ iam:
     oidc-issuer: ${IAM_REGISTRATION_OIDC_ISSUER:https://example.org}
     saml-entity-id: ${IAM_REGISTRATION_SAML_ENTITY_ID:urn:example}
     add-nickname-as-attribute: ${IAM_ADD_NICKNAME_AS_ATTRIBUTE:false}
+    fields:
+      notes:
+        field-behaviour: ${IAM_REGISTRATION_NOTES_FIELD_BEHAVIOUR:mandatory}
```
enricovianello, line 159: If we add this env variable we should add also the others: IAM_REGISTRATION_GIVEN_NAME_BEHAVIOUR, etc. I think we can avoid adding these lines and keep the default value as MANDATORY on this line https://github.com/indigo-iam/iam/pull/782/files#diff-eb3a40a655172a85a4ebaccf2dba240d9b4f294761e61a5631af9892aa95a49fR206

iam-login-service/src/test/java/it/infn/mw/iam/test/registration/RegistrationUnprivilegedTests.java
```diff
@@ -193,7 +205,7 @@ public void testUsernameAlreadyTaken() throws Exception {
   }
 
   @Test
-  public void testCreateRequestWithoutNotes() throws Exception {
+  public void testCreateRequestWithMandatoryNotesField() throws Exception {
```
enricovianello, line 208: We need also to test when notes is not mandatory and we can do this on other classes that has @SpringBootTest(properties = { "iam.registration.fields.notes.field-behaviour=optional" }) on top or @SpringBootTest(properties = { "iam.registration.fields.notes.field-behaviour=hidden" }). This last value can be tested in the same way optional is. You can also test them by testing only the Validators or the Service, without all the mvc part.

enricovianello: but if you prefer the mvc tests are ok :)

enricovianello: but you have already done something similar here https://github.com/indigo-iam/iam/pull/782/files#diff-015aa3049cced12a11b63113340fe4f005b8a504559522e34b3818d667e6043aR106 (but optional tests are missing, or am i missing them?)

enricovianello (changes requested):
I started the review and mainly the big change suggested is to use an implementation of the RequestValidatorService instead of adding the validation logic into the DefaultRegistrationRequestService. I know I'm a bit annoying but I'd prefer to keep things separated, having something similar already done for CERN purposes.

iam-login-service/src/main/java/it/infn/mw/iam/registration/DefaultRegistrationRequestService.java
```diff
@@ -181,9 +183,22 @@ private void createAupSignatureForAccountIfNeeded(IamAccount account) {
...
+    if (fields.containsKey("notes")) {
+      RegistrationFieldProperties notesFieldAttribute = fields.get("notes");
+
+      if ("mandatory".equalsIgnoreCase(notesFieldAttribute.getFieldBehaviour().name())) {
+        notesSanityChecks(dto.getNotes());
+      }
+    }
 
     if (!isNull(validationService)) {
       RegistrationRequestValidationResult result =
```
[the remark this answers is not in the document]
Sae126V: Cool. Make sense @enricovianello . Updated that in 9d06dd3a089c4172d02077a4540b1c783a355130

iam-login-service/src/main/java/it/infn/mw/iam/registration/RegistrationRequestService.java
```diff
@@ -16,15 +16,18 @@
...
 import it.infn.mw.iam.authn.ExternalAuthenticationRegistrationInfo;
+import it.infn.mw.iam.config.IamProperties.RegistrationFieldProperties;
 import it.infn.mw.iam.core.IamRegistrationRequestStatus;
 
 public interface RegistrationRequestService {
 
   RegistrationRequestDto createRequest(RegistrationRequestDto request,
-      Optional<ExternalAuthenticationRegistrationInfo> extAuthnInfo);
+      Optional<ExternalAuthenticationRegistrationInfo> extAuthnInfo,
+      Map<String, RegistrationFieldProperties> fields);
```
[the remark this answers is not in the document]
Sae126V: Done 👍 9d06dd3a089c4172d02077a4540b1c783a355130

iam-login-service/src/test/java/it/infn/mw/iam/test/registration/RegistrationUnprivilegedTests.java
Sae126V: Removed that Enrico 👍 . And implemented that in "you can also test them by testing only the Validators or the Service" part. 9d06dd3a089c4172d02077a4540b1c783a355130

iam-login-service/src/main/webapp/resources/iam/apps/registration/registration.controller.js
```diff
@@ -197,10 +266,53 @@ function RegistrationController(
...
+    if (field.$error.required) {
+        if (fieldName === 'notes') {
+            return 'Please provide a reason for your registration request';
+        }
+        return 'Please provide ' + fieldInfo.articleToUse + ' ' + fieldInfo.label.toLowerCase();
+    } else if (field.$error.minlength) {
+        return fieldInfo.label + ' must be at least ' + fieldInfo.minlength + ' characters long';
+    } else if (field.$error.email) {
+        return 'This is not a valid email';
+    } else if (field.$error.emailAvailable) {
```
rmiccoli, line 310: I think these two validators (for email and username) are not working correctly. I am able to enter an existing username or email without getting any errors. Only after clicking on ‘Register’ do I get an empty red box at the bottom of the page and the exception is raised saying that the email or username already exists.

rmiccoli:
Note that recently this check https://github.com/Sae126V/iam/blob/issue-759-registration-ui/iam-login-service/src/main/java/it/infn/mw/iam/registration/RegistrationRequestDto.java#L97 has been added for the Notes field and I saw it is already merged in this PR. I think this check should be removed otherwise the behavior of the _optional_ or _hidden_ Notes field does not work as it should.

PR #790, opened 2024-06-14
Create a statistical anonymous endpoint

rmiccoli:


iam-login-service/src/main/java/it/infn/mw/iam/core/UserCount.java
```diff
@@ -0,0 +1,33 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package it.infn.mw.iam.core;
+
+public class UserCount {
+
+  private String numberOfUsers;
```
giacomini, line 20: Since it's a _count_, wouldn't it be better to keep the `int` (or `long`) here and do the conversion to string somewhere else?

rmiccoli: Yes indeed, and probably no need to convert it to string either

iam-login-service/src/main/java/it/infn/mw/iam/core/UserCount.java
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package it.infn.mw.iam.core;
+
+public class UserCount {
```
enricovianello, line 18: In order to anticipate the evolution of this response content, I'd prefer to rename this class to something like StatisticalEndpointResponse

enricovianello (changes requested):
Just a renaming :)

enricovianello (approved):

PR #791, opened 2024-06-14
Ensure subject DN is stored into db with the expected format (RFC2253)

enricovianello:
https://datatracker.ietf.org/doc/html/rfc2253
The target of this PR is ensuring that each time a x509 certificate is added (via SCIM PATCH) or linked (during login phase) its subject and issuer are stored in the same way, following RFC 2253. Also the search logic has to be able to "normalize" a received subject to the same RFC, in order to avoid case-insensitive problems.


enricovianello:
Superseded by other PR already released.

PR #793, opened 2024-06-19
Add groups enrollment logic to be applied after users registration

garaimanoj:
Properties to be added in application.yml
```
iam:
  registration:
    default-groups:
      - name: /atlas
        enrollment: INSERT
      - name: /privileged
        enrollment: GROUP_REQUEST
```

When setting up new user accounts, administrators often need to determine which group these accounts should belong to by default. Here’s what you need to know:

**name (Default Group):**

- The term “name” represents the specific group that serves as the default destination for new accounts.

- When a user account is created, it automatically becomes a member of this default group unless specified otherwise.

**enrollment Types:**

 There are two enrollment types that determine how new accounts join the default group:

- INSERT Enrollment:  With this type, the system swiftly adds the new account to the default group during account creation.
No additional steps or user interaction are required—the account slips seamlessly into the group.

         

- GROUP_REQUEST Enrollment **(Not implemented)**:  Here, the process is a bit more interactive. New request to join the group will be sent to the admin. 

**Current Functionality:**
As of now, the system has implemented the INSERT enrollment type.
When creating a new account, it automatically becomes part of the default group (“name”) if that group exists.


iam-login-service/src/main/resources/application.yml
```diff
@@ -154,6 +154,11 @@ iam:
     oidc-issuer: ${IAM_REGISTRATION_OIDC_ISSUER:https://example.org}
     saml-entity-id: ${IAM_REGISTRATION_SAML_ENTITY_ID:urn:example}
     add-nickname-as-attribute: ${IAM_ADD_NICKNAME_AS_ATTRIBUTE:false}
+    default-groups:
```
enricovianello, line 157: These settings must be removed because this application.yml is the default configuration file. This configuration feature must be properly added to IAM documentation.

iam-login-service/src/test/java/it/infn/mw/iam/test/service/IamAccountServiceTests.java
```diff
@@ -843,4 +860,40 @@ public void testSetEndTimeWorksForNonNullDate() {
...
+  @Test
+  public void testNoDefaultGroupsAddedWhenDefaultGroupsNotGiven() {
+    IamAccount account = cloneAccount(CICCIO_ACCOUNT);
+
+    account = accountService.createAccount(account);
+
+    Optional<IamAccountGroupMembership> groupMembershipOptional = account.getGroups().stream().findFirst();
+    assertFalse(groupMembershipOptional.isPresent());
+  }
+
```
enricovianello, line 898: This is ok, it's a good mocking. Just for your knowledge (if needed) you can also write mvc tests and set properties by using a strategy like this https://github.com/indigo-iam/iam/blob/master/iam-login-service/src/test/java/it/infn/mw/iam/test/scim/me/ScimMeFullResponseEndpointTests.java#L44

iam-login-service/src/main/java/it/infn/mw/iam/core/user/DefaultIamAccountService.java
```diff
@@ -177,10 +185,21 @@ public IamAccount createAccount(IamAccount account) {
...
         "Account created for user " + account.getUsername()));
-
+    
+    addToDefaultGroups(account);
     return account;
   }
 
+  private void addToDefaultGroups(IamAccount account) {
+    List<Map<String, String>> defaultGroups = iamProperties.getRegistration().getDefaultGroups();
+    if(defaultGroups != null){
```
enricovianello, line 195: it seems this lines are not well formatted (no spaces between if and parenthesis e.g.)

iam-login-service/src/main/java/it/infn/mw/iam/config/IamProperties.java
```diff
@@ -232,6 +232,8 @@ public static class RegistrationProperties {
 
     Map<String, RegistrationFieldProperties> fields;
 
+    List<Map<String, String>> defaultGroups;
```
enricovianello, line 235: I'd prefer to define an inner class DefaultGroup with name and enrollment fields

enricovianello: it helps to define default value for enrollment (INSERT) and you can specify a NotEmpty validation for name. Of course this defaultGroups list must remain nullable as it is.

enricovianello (changes requested):
The main changement is related to addind a DefaultGroup properties class, such as all the others already defined into IamProperties

iam-login-service/src/main/resources/application.yml
garaimanoj: Done

iam-login-service/src/main/java/it/infn/mw/iam/core/user/DefaultIamAccountService.java
garaimanoj: Done

iam-login-service/src/main/java/it/infn/mw/iam/config/IamProperties.java
garaimanoj: I have added the default `INSERT` value. Do you have any example of NotEmpty validation please?

iam-login-service/src/main/java/it/infn/mw/iam/core/user/DefaultIamAccountService.java
```diff
@@ -177,10 +185,22 @@ public IamAccount createAccount(IamAccount account) {
...
         "Account created for user " + account.getUsername()));
-
+    
+    addToDefaultGroups(account);
     return account;
   }
 
+  private void addToDefaultGroups(IamAccount account) {
+    List<DefaultGroup> defaultGroups = iamProperties.getRegistration().getDefaultGroups();
+    if (defaultGroups != null) {
```
enricovianello, line 195: We can use also Objects.nonNull(defaultGroups)

enricovianello (changes requested):
Just a minor cosmetic fix requested. Another thing we could add to this PR description the example of the application.yml configuration required (e.g. the part we removed).
This will make easier to understand it when reading new features from release notes.

iam-login-service/src/main/java/it/infn/mw/iam/config/IamProperties.java
enricovianello: I saw validation at this level is not done so we can live without it at this point. The existence of the group name in implicitly done later

PR #795, opened 2024-06-21
Fix client track-last-used setting location in .yaml

enricovianello:


enricovianello:
We should also align the default value to true inside `IamProperties` class (`trackLastUsed` field)

darcato (approved):
The changes fix the default behaviour

darcato:
> We should also align the default value to true inside `IamProperties` class (`trackLastUsed` field)

Done

PR #802, opened 2024-07-02
Send an email when client status changes

rmiccoli:
Emails are sent to client contacts and client owners, if defined.


iam-login-service/src/main/java/it/infn/mw/iam/api/client/management/service/DefaultClientManagementService.java
```diff
@@ -139,10 +145,21 @@ public void deleteClientByClientId(String clientId) {
   public void updateClientStatus(String clientId, boolean status, String userId) {
 
     ClientDetailsEntity client = clientService.findClientByClientId(clientId)
-        .orElseThrow(ClientSuppliers.clientNotFound(clientId));
+      .orElseThrow(ClientSuppliers.clientNotFound(clientId));
     client = clientService.updateClientStatus(client, status, userId);
-    String message = "Client " + (status?"enabled":"disabled");
+    String message = "Client " + (status ? "enabled" : "disabled");
     eventPublisher.publishEvent(new ClientStatusChangedEvent(this, client, message));
+    if (!client.getContacts().isEmpty()) {
```
enricovianello, line 152: What happens if owner's email is also within client's contacts? Probably we could build a set of emails (no duplicates) from client contacts and owners and this can be done through a unique notificationFactory.createClientStatusChangedMessageFor(ClientDetailsEntity client) instead of 2 methods?

iam-login-service/src/main/java/it/infn/mw/iam/notification/NotificationFactory.java
```diff
@@ -39,4 +41,9 @@ public interface NotificationFactory {
   IamEmailNotification createGroupMembershipApprovedMessage(IamGroupRequest groupRequest);
 
   IamEmailNotification createGroupMembershipRejectedMessage(IamGroupRequest groupRequest);
+
+  IamEmailNotification createClientStatusChangedMessage(ClientDetailsEntity client);
+
```
enricovianello, line 46: I'd prefer (as suggested above) a unique method that receives client object (owners and contacts can be retrieved from it)

enricovianello (changes requested):
some optimizations :)

iam-login-service/src/main/resources/email-templates/clientStatusChanged.ftl
```diff
@@ -0,0 +1,5 @@
+Dear user,
+
+this mail is to inform you that client ${clientName} with id ${clientId} has been <#if clientStatus>ACTIVATED<#else>SUSPENDED</#if>.
```
enricovianello, line 3: clientStatus --> isClientActive

rmiccoli:
> some optimizations :)

Done :)

enricovianello (approved):

PR #810, opened 2024-07-19
Specify the organization name in the email subject

rmiccoli:
This PR ensures that all the subjects of the various emails have a prefix, equal to `[VO name + IAM]`, e.g.
in case of ATLAS VO all email subjects will be `[atlas IAM] ...`


darcato:
Maybe it's better to add the VO name to the subject inside the `createMessage` method, so that it applies to all the messages.

rmiccoli:
> Maybe it's better to add the VO name to the subject inside the `createMessage` method, so that it applies to all the messages.

Done :)

enricovianello (approved):

enricovianello:
Could you improve this PR description? @rmiccoli Thanks :) Then, we can surely sq+merge it

PR #811, opened 2024-07-22
AUP exemption

garaimanoj:


iam-login-service/src/test/java/it/infn/mw/iam/test/notification/RegistrationFlowNotificationTests.java
```diff
@@ -328,4 +345,73 @@ public void testRejectFlowMotivationNotifications() throws Exception {
 
   }
 
+  @Test
```
garaimanoj, line 348: If I put this tests in a new file, tests are not run by GitHub and test coverage is showing 40%

rmiccoli: > If I put this tests in a new file, tests are not run by GitHub and test coverage is showing 40%

The name of the new file ended with 'Tests'? If not, it is ignored

iam-login-service/src/test/java/it/infn/mw/iam/test/notification/ServiceAccountNotificationTest.java
```diff
@@ -0,0 +1,167 @@
...
+    @Autowired
+    private IamAccountRepository iamAccountRepo;
+
+    @Autowired
+    private IamEmailNotificationRepository notificationRepo;
+
+    private MockMvc mvc;
+
+    @Before
+    public void setUp() throws InterruptedException {
```
rmiccoli, line 88: Remove InterruptedException as also suggested by sonarcloud

iam-login-service/src/test/java/it/infn/mw/iam/test/notification/ServiceAccountNotificationTest.java
```diff
@@ -0,0 +1,167 @@
...
+        mvc = MockMvcBuilders.webAppContextSetup(context).alwaysDo(log()).apply(springSecurity()).build();
+    }
+
+    @After
+    public void tearDown() throws InterruptedException {
```
rmiccoli, line 93: The same as above

iam-persistence/src/main/java/it/infn/mw/iam/persistence/repository/IamAupSignatureRepositoryImpl.java
```diff
@@ -68,7 +68,9 @@ private void deleteSignature(IamAccount account) {
   @Override
   public IamAupSignature createSignatureForAccount(IamAup aup, IamAccount account,
       Date currentTime) {
-
+    if (account.isServiceAccount()) {
+      throw new IamAupSignatureUpdateError(account, "As user '%s' is a service account, AUP signature operation not allowed");
```
rmiccoli, line 72: To avoid repetition, I'd define the message directly in the IamAupSignatureUpdateError constructor

iam-common/src/main/java/it/infn/mw/iam/service/aup/DefaultAupSignatureCheckService.java
```diff
@@ -53,7 +53,7 @@ public boolean needsAupSignature(IamAccount account) {
 
     Date now = new Date(timeProvider.currentTimeMillis());
 
-    if (!aup.isPresent()) {
+    if (!aup.isPresent() || account.isServiceAccount()) {
```
rmiccoli, line 56: I'd separate the two conditions to have two different log messages. In case of service accounts a message like this "_AUP signature not needed for account '{}': Account is a service account_".

iam-persistence/src/main/resources/db/migration/h2/V106__add_service_account.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE iam_account ADD COLUMN service_account BOOLEAN DEFAULT false;
+
+UPDATE iam_account SET service_account = false;
```
rmiccoli, line 3: Is the update necessary? I think the first statement is sufficient

iam-login-service/src/test/java/it/infn/mw/iam/test/notification/ServiceAccountNotificationTest.java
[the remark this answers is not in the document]
garaimanoj: Done

iam-common/src/main/java/it/infn/mw/iam/service/aup/DefaultAupSignatureCheckService.java
garaimanoj: New if condition added

iam-persistence/src/main/java/it/infn/mw/iam/persistence/repository/IamAupSignatureRepositoryImpl.java
garaimanoj: Done

iam-login-service/src/test/java/it/infn/mw/iam/test/notification/ServiceAccountNotificationTest.java
garaimanoj: Removed

iam-login-service/src/test/java/it/infn/mw/iam/test/notification/ServiceAccountNotificationTest.java
garaimanoj: Removed

iam-persistence/src/main/resources/db/migration/h2/V106__add_service_account.sql
garaimanoj: yes, second line removed

rmiccoli (commented):
In general, the PR looks good to me.

I have only some comments:
* would it be better to handle the IamAupSignatureUpdateError exception and print the error message? now only the status code 405 is shown
* if the user is set up as a service account after having already signed the AUP, does it make sense to set the AUP signature time to null? this should not even trigger the AUP expiry reminder emails

enricovianello:
> * would it be better to handle the IamAupSignatureUpdateError exception and print the error message? now only the status code 405 is shown

I agree, we could add an ExceptionHandler like this in the same controller this https://github.com/indigo-iam/iam/pull/811/files#diff-a8bc81ad573d8d00f0b71699cc5862722f1bc38abdd3c9cd03e206639a9f904aR205-R209

> * if the user is set up as a service account after having already signed the AUP, does it make sense to set the AUP signature time to null? this should not even trigger the AUP expiry reminder emails

It makes sense to me. In my opinion when the administrator is prompted with the message of "service account creation" it should be aware of the fact that this account will lose the signatures and it will be exempted from them in the future. Again, as soon as the admin is going to switch again from a service account to a "normal" one, we could be very clear in the message by explaining that the user won't have any signature so he can decide to sign it on behalf in case.
I also prefer that the AUP reminder logic is updated with a filter that excludes service account when it looks for the accounts that have to receive the email.

garaimanoj:
> In my opinion when the administrator is prompted with the message of "service account creation" it should be aware of the fact that this account will lose the signatures and it will be exempted from them in the future. Again, as soon as the admin is going to switch again from a service account to a "normal" one, we could be very clear in the message by explaining that the user won't have any signature so he can decide to sign it on behalf in case.

We had a discussion regarding deletion of existing AUP signature. During that @enricovianello mentioned that if Admin by mistake set the account as `service account` then we will loose the AUP signature. So we can keep the signature as is and when account will be back as `normal account` we can have all the signature data as is.

I think checking the service account status during AUP reminder is a good idea. If all agree I could add that additional check.

rmiccoli:
It is true that when the account is set up as a service account, the signature of the aup is skipped (if it is defined), but what if it accidentally goes to the `/iam/aup/sign` endpoint and accepts the aup? the exception is thrown at this time, I don't know if the ExceptionHandler should be added here as well https://github.com/indigo-iam/iam/blob/bddf581818fd66442bc86a0a951de2bf6ae1c0cd/iam-login-service/src/main/java/it/infn/mw/iam/api/aup/AupSignaturePageController.java#L99

garaimanoj:
> When the account is set up as a service account, the signature of the aup is skipped (if it is defined), but what if it accidentally goes to the `/iam/aup/sign` endpoint and accepts the aup? the exception is thrown at this time, I don't know if the ExceptionHandler should be added here as well
> 
> https://github.com/indigo-iam/iam/blob/bddf581818fd66442bc86a0a951de2bf6ae1c0cd/iam-login-service/src/main/java/it/infn/mw/iam/api/aup/AupSignaturePageController.java#L99

I am going to add the exception handler and a test.

iam-login-service/src/main/java/it/infn/mw/iam/api/aup/AupSignaturePageController.java
```diff
@@ -126,6 +131,12 @@ public ModelAndView signAup(HttpServletRequest request, HttpServletResponse resp
 
     return new ModelAndView("redirect:/dashboard");
   }
+
+  @ResponseStatus(value = HttpStatus.METHOD_NOT_ALLOWED)
+  @ExceptionHandler(IamAupSignatureUpdateError.class)
+  public ErrorDTO aupSignatureUpdateError(Exception ex) {
```
rmiccoli, line 137: In this case, it would be better to return the error message as a JSON response?
```
@ExceptionHandler(IamAupSignatureUpdateError.class)
  public ResponseEntity<ErrorDTO> aupSignatureUpdateError(Exception ex) {
      ErrorDTO errorResponse = ErrorDTO.fromString(ex.getMessage());
      return new ResponseEntity<>(errorResponse, HttpStatus.METHOD_NOT_ALLOWED);
  }
```

garaimanoj: Thanks. Done.

rmiccoli (approved):

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/model/ScimUser.java
```diff
@@ -93,7 +94,8 @@ private ScimUser(@JsonProperty("id") String id, @JsonProperty("externalId") Stri
       @JsonProperty("userType") String userType,
       @JsonProperty("preferredLanguage") String preferredLanguage,
       @JsonProperty("locale") String locale, @JsonProperty("timezone") String timezone,
-      @JsonProperty("active") Boolean active, @JsonProperty("emails") List<ScimEmail> emails,
+      @JsonProperty("active") Boolean active, @JsonProperty("serviceAccount") Boolean serviceAccount,
```
enricovianello, line 97: Unfortunately `serviceAccount` is not a valid SCIM User field. We need to move this inside ScimIndigoUser object which uses the SCIM extension mechanism

enricovianello: https://www.rfc-editor.org/rfc/rfc7643.html#section-4.1

enricovianello (changes requested):
If we add things to SCIM user that are not in the SCIM protocol we need to add them inside ScimIndigoUser extension object

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/model/ScimUser.java
garaimanoj: Done

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/model/ScimUser.java
```diff
@@ -216,6 +217,11 @@ public Boolean getActive() {
     return active;
   }
 
+  public Boolean getServiceAccount() {
```
rmiccoli, line 220: I think we don't need it here anymore, otherwise we get _service account_ claim also in `ScimUser` object.

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/updater/factory/DefaultAccountUpdaterFactory.java
```diff
@@ -208,6 +208,10 @@ private void prepareReplacers(List<AccountUpdater> updaters, ScimUser user, IamA
     addUpdater(updaters, Objects::nonNull, user::getPassword, replace::password);
     addUpdater(updaters, Objects::nonNull, user::getActive, replace::active);
 
+    if (user.hasServiceAccountStatus()) {
+      addUpdater(updaters, Objects::nonNull, user::getServiceAccount, replace::serviceAccount);
```
rmiccoli, line 212: here it becomes `user.getIndigoUser()::getServiceAccount`

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/converter/UserConverter.java
```diff
@@ -91,6 +91,10 @@ public IamAccount entityFromDto(ScimUser scimUser) {
       account.setActive(scimUser.getActive());
     }
 
+    if (scimUser.getServiceAccount() != null) {
```
rmiccoli, line 94: it becomes
```
if (scimUser.hasServiceAccountStatus()) {
      account.setServiceAccount(scimUser.getIndigoUser().getServiceAccount());
}
 ```

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/model/ScimUser.java
garaimanoj: Thanks. It makes sense.

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/updater/factory/DefaultAccountUpdaterFactory.java
garaimanoj: Done

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/converter/UserConverter.java
garaimanoj: Done

rmiccoli (approved):

enricovianello (approved):

PR #817, opened 2024-07-25
Assign oidc-agent clients to whom approved it

enricovianello:
If a user approved a client with name prefix _oidc-agent:_, it will be automatically linked to the user.

This should have been enabled for the device and authorization code flows.


rmiccoli (changes requested):
The test that fails is due to a constraint on a column of a mysql table (not h2). Consider correcting/eliminating the test if not necessary

federicaagostini:
> The test that fails is due to a constraint on a column of a mysql table (not h2). Consider correcting/eliminating the test if not necessary

Test removed, as it will never be reproducible in production due to mysql constraint.

rmiccoli (approved):

federicaagostini:
Superseded by https://github.com/indigo-iam/iam/pull/821

PR #821, opened 2024-07-31
Create an approved site when using device code flow

federicaagostini:
The device code flow did not imply the creation of an approved site when the user give the consent to the client, as the authorization code flow does.
This PR adds an approved site using the same logic as the authorization code flow. Moreover, the scope policy filter has been added here for both the authorization code flow and device code flow.

This required to move (and cleanup) MitreID classes related to the device code flow into IAM.


iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/IamDeviceEndpointController.java
```diff
@@ -257,6 +261,10 @@ public String approveDevice(@RequestParam("user_code") String userCode,
 
     model.put("approved", true);
 
+    Date timeout = null;
+    approvedSiteService.createApprovedSite(client.getClientId(), auth.getName(), timeout,
```
rmiccoli, line 265: The problem here is the case where there is a scope policy. Since it is evaluated after this step, in the approved sites we will see also the filtered scope - and in fact we see it on the consent page and actually approve it, even though it is then removed from the scope policy filter in the access token. 
In generale, this mitre code needs to be improved later!

rmiccoli (approved):
Fix only test

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/IamDeviceEndpointController.java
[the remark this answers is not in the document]
rmiccoli: > The problem here is the case where there is a scope policy. Since it is evaluated after this step, in the approved sites we will see also the filtered scope - and in fact we see it on the consent page and actually approve it, even though it is then removed from the scope policy filter in the access token. In general, this mitre code needs to be improved later, but as a start I think it might be fine like this!

Scope policy filter added before the device consent page!

federicaagostini:
From Gabriel Zachmann: _the client should only be linked to the user account if it currently is not linked to any user or at least not for public clients._

iam-login-service/src/test/java/it/infn/mw/iam/test/oauth/scope/pdp/ScopePolicyFilteringDeviceCode.java
rmiccoli, line 1: Rename this test to ScopePolicyFilteringDeviceCodeTests.java

federicaagostini:
Superseded by https://github.com/indigo-iam/iam/pull/908.

PR #834, opened 2024-08-22
Fix AUP signature validity

rmiccoli:
The custom validator for AUP signature validity and AUP reminders, introduced within IAM v1.10.0, checks that the value of the AUP reminders is less than the AUP signature validity. This PR fixes the case AUP signature validity is equal to 0. In particular:
* if AUP signature validity is set to 0, AUP reminders are set to an empty string
* when the AUP signature validity is 0, AUP reminders are also hidden from the interface

### DIFF
--- a/iam-login-service/src/main/java/it/infn/mw/iam/api/aup/model/AupRemindersAndSignatureValidator.java
+++ b/iam-login-service/src/main/java/it/infn/mw/iam/api/aup/model/AupRemindersAndSignatureValidator.java
@@ -63,50 +63,57 @@ public class AupRemindersAndSignatureValidator
       return true;
     }
 
-    if (aupRemindersInDays != null) {
-      try {
-        List<Integer> numbers = Arrays.stream(aupRemindersInDays.split(","))
-          .map(String::trim)
-          .map(Integer::parseInt)
-          .collect(Collectors.toList());
+    if (aupRemindersInDays == null) {
+      context.disableDefaultConstraintViolation();
+      context.buildConstraintViolationWithTemplate(
+          "Invalid AUP: aupRemindersInDays must be set when signatureValidityInDays is greater than 0")
+        .addConstraintViolation();
+      return false;
+    }
 
-        if (numbers.stream().anyMatch(i -> i <= 0)) {
-          context.disableDefaultConstraintViolation();
-          context
-            .buildConstraintViolationWithTemplate(
-                "Invalid AUP: zero or negative values for reminders are not allowed")
-            .addConstraintViolation();
-          return false;
-        }
+    try {
+      List<Integer> numbers = Arrays.stream(aupRemindersInDays.split(","))
+        .map(String::trim)
+        .map(Integer::parseInt)
+        .collect(Collectors.toList());
 
-        if (numbers.stream().anyMatch(i -> i >= signatureValidityInDays)) {
-          context.disableDefaultConstraintViolation();
-          context
-            .buildConstraintViolationWithTemplate(
-                "Invalid AUP: aupRemindersInDays must be smaller than signatureValidityInDays")
-            .addConstraintViolation();
-          return false;
-        }
-
-        Set<Integer> uniqueNumbers = new HashSet<>(numbers);
-        if (uniqueNumbers.size() != numbers.size()) {
-          context.disableDefaultConstraintViolation();
-          context
-            .buildConstraintViolationWithTemplate(
-                "Invalid AUP: duplicate values for reminders are not allowed")
-            .addConstraintViolation();
-          return false;
-        }
-
-        return true;
-      } catch (NumberFormatException e) {
+      if (numbers.stream().anyMatch(i -> i <= 0)) {
         context.disableDefaultConstraintViolation();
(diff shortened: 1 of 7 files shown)


enricovianello (approved):

iam-login-service/src/main/java/it/infn/mw/iam/api/aup/model/AupRemindersAndSignatureValidator.java
```diff
@@ -50,6 +50,11 @@ public boolean isValid(AupDTO value, ConstraintValidatorContext context) {
       return false;
     }
 
+    if (signatureValidityInDays == 0) {
+      value.setAupRemindersInDays("30,15,1");
+      return true;
+  }
```
enricovianello, line 56: indentation problem

iam-login-service/src/main/java/it/infn/mw/iam/api/aup/model/AupRemindersAndSignatureValidator.java
```diff
@@ -50,6 +50,11 @@ public boolean isValid(AupDTO value, ConstraintValidatorContext context) {
       return false;
     }
 
+    if (signatureValidityInDays == 0) {
+      value.setAupRemindersInDays("30,15,1");
```
enricovianello, line 54: can we move the set out from validation?

enricovianello: In alternative, we can both clear reminders value from dashboard PUT request in case validity in days is = 0 and return bad request in case the AupDTO contains validity in days = 0 and a value for reminders.

iam-login-service/src/main/java/it/infn/mw/iam/api/aup/model/AupRemindersAndSignatureValidator.java
```diff
@@ -50,55 +51,62 @@ public boolean isValid(AupDTO value, ConstraintValidatorContext context) {
...
 
-      if (numbers.stream().anyMatch(i -> i >= signatureValidityInDays)) {
-        context.disableDefaultConstraintViolation();
-        context
-          .buildConstraintViolationWithTemplate(
-              "Invalid AUP: aupRemindersInDays must be smaller than signatureValidityInDays")
-          .addConstraintViolation();
-        return false;
-      }
+    if (aupRemindersInDays != null) {
```
enricovianello, line 66: at this point we're sure that signatureValidityInDays is greater than zero then we expect a value for aupRemindersInDays

I think we need something like if aupRemindersInDays is null add a constraint "Invalid AUP: aupRemindersInDays must be set when signatureValidityInDays is greater than zero"
We don't need to check empty value because probably with an empty string the code below will raise an exception

iam-login-service/src/main/webapp/resources/iam/apps/dashboard-app/components/aup/aup.create.dialog.html
```diff
@@ -22,19 +22,23 @@ <h3 class="modal-title">
...
     <form name="createAup">
-        <div class="form-group" ng-class="{'has-error': $ctrl.error || createAup.url.$dirty && !createAup.url.$valid}">
+        <div class="form-group has-error" ng-if="$ctrl.error">
+            <span class="help-block">
+                {{ $ctrl.error }}
+            </span>
+        </div>
+        <div class="form-group">
             <label>Acceptable Usage Policy URL</label>
-            <input name="url" class="form-control" type="url" value="http://" ng-model="$ctrl.aupVal.url" required="true">
```
enricovianello, line 27: Why type="url" was a problem?

iam-login-service/src/main/webapp/resources/iam/apps/dashboard-app/components/aup/aup.edit.dialog.html
```diff
@@ -22,9 +22,14 @@ <h3 class="modal-title">
...
     <form name="createAup">
-        <div class="form-group" ng-class="{'has-error': $ctrl.error || createAup.url.$dirty && !createAup.url.$valid}">
+        <div class="form-group has-error" ng-if="$ctrl.error">
+            <span class="help-block">
+                {{ $ctrl.error }}
+            </span>
+        </div>
+        <div class="form-group">
             <label>Acceptable Usage Policy URL</label>
-            <input name="url" class="form-control" type="url" value="http://" ng-model="$ctrl.aupVal.url"
```
enricovianello, line 27: same question as above

iam-login-service/src/main/webapp/resources/iam/apps/dashboard-app/components/aup/aup.create.dialog.html
rmiccoli: Because `type="url"` expects a URL and and if we add a string in the form it is ignored and thus we get a wrong error from the url validator, i.e. I`nvalid AUP: the AUP URL cannot be blank` instead of `Invalid AUP: the AUP URL is not valid`

enricovianello (approved):

PR #840, opened 2024-09-04
Allow use of JDBC session on separate DB

### LINKED ISSUE #809: JDBC for Cacheing/session storage
**Describe The Problem To Be Solved**
I am trying to deploy a distributed IAM and running a HA redis infrastructure will increase the complexity and chances of failure of the system. Thus, I propose adding JDBC support for external session storage.

**Solution suggestion**
* Details of the technical implementation
  - Add `spring-jdbc-session` dependency 
  - review and disable any caching strategy that is not suitable for SQL
  - (Optional) support a separate datasource(database) for session storage, most HA MySQL options will benefit from having caching at a separate DB as most IAM data are written in a low frequency while catching and session are accessed at high frequency.
* Tradeoffs made in design decisions
  - May reduce performance as SQL are in general slower
* Caveats and considerations for the future
  - Addition of Postgres option as it offers faster distributed HA options such as CockroachDB

DonaldChung-HK:
# Description
This change allows the use of JDBC for saving HTTPSession data in a separate database. This is to support our use of Galera DB as our HA backend option. The change will allow use to reduce administrative burden and attack surface via not needing to run a cross data centre Redis. setup.

The code should only activate if `jdbc` spring session is used.

Fixes #809


iam-login-service/src/main/java/it/infn/mw/iam/config/DataSourceConfig.java
```diff
@@ -0,0 +1,50 @@
+package it.infn.mw.iam.config;
```
enricovianello, line 1: This is my usual fault :) Forgetting the license header! you can copy it from other Java classes or from LICENSE file. I know we should upgrade it. We'll do it in one shot for all the files soon.

DonaldChung-HK: I have added the header, thanks for pointing this out.

DonaldChung-HK:
I have added the missing new line at `iam-login-service/src/main/resources/application-jdbc-session.yml` so the licence check should pass now

rmiccoli:
I think we don't need `@Primary` and `@Qualifier` annotations since Spring will manage the different data sources based on the condition set. So,
* when `spring.session.store-type=jdbc` is enabled, both primaryDataSource and sessionDataSource are created, and Spring Session uses _session_ database for storage persistence;
* when `spring.session.store-type=jdbc` is not set, only primaryDataSource is created and used for everything.

iam-login-service/src/main/java/it/infn/mw/iam/config/DataSourceConfig.java
```diff
@@ -0,0 +1,65 @@
...
+        return dataSourceProperties
+                .initializeDataSourceBuilder()
+                .type(HikariDataSource.class)
+                .build();
+    }
+
+    @Bean
+    @ConfigurationProperties("session.datasource")
+    @Qualifier("sessionDataSourceProperties")
+    @ConditionalOnProperty(name = "spring.session.store-type", havingValue = "jdbc")
```
rmiccoli, line 51: Also this annotation should be unnecessary here

DonaldChung-HK:
Hi, I think it does need to `@Qualifier` since otherwise there will be 2x non unique beans for datasource, one is the autoconfigured one and another one is this customer one?

DonaldChung-HK:
Hi, 
The test I have above is failing due to MySQL too many connections at the `OidcExternalAuthenticationTests`. It is strange cause in local development environment, all tests pass but it seems that in local environment, it uses `h2` instead of `MySQL`. Could you please investigate?

rmiccoli:
Replaced by https://github.com/indigo-iam/iam/pull/898

PR #845, opened 2024-09-20
Fix X509 certificate linking

rmiccoli:


iam-login-service/src/main/java/it/infn/mw/iam/api/account_linking/DefaultAccountLinkingService.java
```diff
@@ -140,56 +142,49 @@ public void linkX509Certificate(Principal authenticatedUser,
...
-      linkedCert.ifPresent(c -> {
-        c.setCertificate(x509Credential.getCertificateChainPemString());
-        c.setLastUpdateTime(new Date());
-      });
+    if (linkedCertificate.isPresent()) {
 
+      linkedCertificate.get().setCertificate(x509Credential.getCertificateChainPemString());
+      linkedCertificate.get().setLastUpdateTime(new Date());
+      certificateRepository.save(linkedCertificate.get());
+      userAccount.getX509Certificates().add(linkedCertificate.get());
```
enricovianello, line 163: Set.add insert the element only if is not contained and it uses equals method to check this. Unluckily our X509Certificate equals method checks only subject and issuer then no PEM certificate. This will mean that the element won't be replaced into that set. Probably it's not a big problem but it's better to remove and then add it again.

enricovianello (approved):

PR #858, opened 2024-10-25
Delete rogue registration requests without sending a reason

garaimanoj:


garaimanoj:
image

rmiccoli:
> image

In my opinion, when the box "Do not send email to the rejected user" is checked, the reason shouldn't be mandatory. Actually it becomes quite useless because it is neither logged nor saved to the database.

iam-login-service/src/main/java/it/infn/mw/iam/registration/DefaultRegistrationRequestService.java
```diff
@@ -358,9 +358,12 @@ private RegistrationRequestDto handleConfirm(IamRegistrationRequest request) {
...
 
   private RegistrationRequestDto handleReject(IamRegistrationRequest request,
-      Optional<String> motivation) {
+      Optional<String> motivation, boolean doNotSendEmail) {
     request.setStatus(REJECTED);
-    notificationFactory.createRequestRejectedMessage(request, motivation);
+    if(!doNotSendEmail){
+      notificationFactory.createRequestRejectedMessage(request, motivation);
+    }
+
```
enricovianello, line 366: I would add the `motivation` to the message published into RegistrationRejectEvent (line 371), knowing it's still required

enricovianello (changes requested):
Easy fix: I would add the motivation to the message published into RegistrationRejectEvent

iam-login-service/src/main/java/it/infn/mw/iam/registration/DefaultRegistrationRequestService.java
garaimanoj: Done

rmiccoli (approved):

enricovianello (approved):

PR #865, opened 2024-11-07
Reset mfa for user

garaimanoj:
A button added to reset/delete MFA for other users by Admin
Label added to show MFA `Active/Disabled`

image


rmiccoli:
> A button added to reset/delete MFA for other users and admin by Admin
> 
> image

Does it make sense for the Admin to reset their own MFA? if they can't log in to their authenticator, they can't log in to IAM, so it should be another Admin who can reset their MFA (or the sys admin acting directly on the db).

garaimanoj:
> > A button added to reset/delete MFA for other users and admin by Admin
> > image
> 
> Does it make sense for the Admin to reset their own MFA? if they can't log in to their authenticator, they can't log in to IAM, so it should be another Admin who can reset their MFA (or the sys admin acting directly on the db).

We were discussing about the scenario. If the Admin login using MFA and after that if there is any issue with the Authenticator App or the phone in that case Admin can reset their own MFA.

Sae126V:
> > > A button added to reset/delete MFA for other users and admin by Admin
> > > image
> > 
> > 
> > Does it make sense for the Admin to reset their own MFA? if they can't log in to their authenticator, they can't log in to IAM, so it should be another Admin who can reset their MFA (or the sys admin acting directly on the db).
> 
> We were discussing about the scenario. If the Admin login using MFA and after that if there is any issue with the Authenticator App or the phone in that case Admin can reset their own MFA.

I am NOT Sure. But, I would prefer admin contacting another Admin who can reset their MFA (or the sys admin acting directly on the db).  We can discuss about this on the call.

garaimanoj:
Some point to discuss
1. Should we use `Reset multi-factor setting` or `Remove multi-factor setting`?
2. When should the status of multi-factor authentication be visible? Currently, the status (Active/Disabled) is always shown to both users and admins

garaimanoj:
> > > > A button added to reset/delete MFA for other users and admin by Admin
> > > > image
> > > 
> > > 
> > > Does it make sense for the Admin to reset their own MFA? if they can't log in to their authenticator, they can't log in to IAM, so it should be another Admin who can reset their MFA (or the sys admin acting directly on the db).
> > 
> > 
> > We were discussing about the scenario. If the Admin login using MFA and after that if there is any issue with the Authenticator App or the phone in that case Admin can reset their own MFA.
> 
> I am NOT Sure. But, I would prefer admin contacting another Admin who can reset their MFA (or the sys admin acting directly on the db). We can discuss about this on the call.

Done. Now admin can not reset own MFA.

rmiccoli:
> Some points to discuss
> 
>     1. Should we use `Reset multi-factor settings` or `Remove multi-factor settings`?
> 
>     2. When should the status of multi-factor authentication be visible? Currently, the status (Active/Disabled) is always shown to both users and admins

We discussed together with @giacomini
1. It's better to have "Disable MFA"
2. It's ok to show the status of MFA always (for both users and admins) but we prefer to have only a (green) tick ✔️ if MFA is enabled and a cross ❌ if disabled, instead of having the green/yellow _Active/Disabled_ box. Then, "Multi-factor authentication" can be abbreviated to "MFA".

Moreover, when MFA is enabled the text of the "Edit multi-factor settings" button can simply become "Disable MFA" and when disabled, it can be "Enable MFA".

rmiccoli (approved):

PR #868, opened 2024-11-13
Handle HEAD requests for /registration/verify/{token} endpoint

rmiccoli:
Other endpoints are ok, `/iam/aup/sign` endpoint fixed in PR https://github.com/indigo-iam/iam/pull/870.


enricovianello:
Merged into next release branch https://github.com/indigo-iam/iam/tree/v1.10.3

Issue #855, opened 2024-10-16
AUP Signing URL should trigger full authentication


hshort:
Going directly to https://IAM/iam/aup/sign does not redirect to the login page, and results in an error (see below). It would be very useful to be able to share this direct link with users, without having to ask them to log in on the homepage and then access the AUP signing page.
 
```
<oauth>
<error_description>Full authentication is required to access this resource</error_description>
<error>unauthorized</error>
</oauth>
```

maarten-litmaath:
As a workaround we could also let the message ask the user to log in and click on the "Re-sign AUP" button...

PR #870, opened 2024-11-14
 Redirect to login page when signing AUP 

federicaagostini:
In case the user is not authenticated, the `/iam/aup/sign` endpoint redirects to the login page.
This should solve issue #855.


enricovianello:
Merged into next release branch https://github.com/indigo-iam/iam/tree/v1.10.3

PR #885, opened 2024-11-27
Prevent the issue of broken SAML login flow

enricovianello:
Fix described here https://www.linkedin.com/pulse/samesite-cookie-infinite-redirections-saml-digvijay-singh/


federicaagostini (approved):

DonaldChung-HK:
To add to this, either the change to the deployed Nginx or the spring section will resolve this issue. The reason the issue is raised is because some site many not have direct access to the nginx configuration.

PR #888, opened 2024-11-29
Add support for resource parameter

federicaagostini:
Following [RFC8707](https://datatracker.ietf.org/doc/html/rfc8707), during an OAuth request the client may indicate a _resource_ parameter, which is the target service  to which access is being requested (MUST be an absolute URI). The value of the resource parameter is returned as _aud_ claim in the JWT.

In case both `audience` and `resource` are provided, only the resource parameter will be included in the token (as _aud_ claim).


iam-login-service/src/main/java/it/infn/mw/iam/core/error/InvalidResourceError.java
```diff
@@ -0,0 +1,32 @@
+/**
+ * Copyright (c) Istituto Nazionale di Fisica Nucleare (INFN). 2016-2021
```
giacomini, line 2: At least for new files, let's use the right year(s). In this case it would just be 2024.

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/IamOAuth2RequestFactory.java
```diff
@@ -164,4 +186,17 @@ public TokenRequest createTokenRequest(Map<String, String> requestParameters,
     return new TokenRequest(requestParameters, clientId, scopes, grantType);
   }
 
+  private boolean isValidUrl(String url) {
```
giacomini, line 189: There is no need to return `boolean` (or anything else), since in case of failure an exception is raised (but I don't know how `forEach` behaves in case of a function not returning anything.

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/IamOAuth2RequestFactory.java
```diff
@@ -164,4 +186,17 @@ public TokenRequest createTokenRequest(Map<String, String> requestParameters,
...
+    try {
+      new URL(url).toURI();
+      return true;
+    } catch (MalformedURLException | URISyntaxException e) {
+      throw new InvalidResourceError("Not a valid URI: " + url);
```
giacomini, line 194: In case of an exception, it's not clear to me at what level the exception is caught; in other words, does the whole request fail? just the setting of the audience? if `resource` is invalid, shall we (re)consider the `aud` parameter instead?

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/IamOAuth2RequestFactory.java
```diff
@@ -164,4 +186,17 @@ public TokenRequest createTokenRequest(Map<String, String> requestParameters,
     return new TokenRequest(requestParameters, clientId, scopes, grantType);
   }
 
+  private boolean isValidUrl(String url) {
+    try {
+      new URL(url).toURI();
```
giacomini, line 191: Does this check that no fragments are present in the url? and what about a query part?

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/IamOAuth2RequestFactory.java
giacomini: can you add some tests for these cases, please?

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/common/BaseAccessTokenBuilder.java
```diff
@@ -123,6 +130,28 @@ protected boolean hasRefreshTokenAudienceRequest(OAuth2Authentication authentica
...
   }
 
+  protected boolean hasValidRefreshTokenResourceRequest(OAuth2Authentication authentication) {
+    TokenRequest refreshTokenRequest = authentication.getOAuth2Request().getRefreshTokenRequest();
+    if (!(isNull(refreshTokenRequest)
+        || isNull(refreshTokenRequest.getRequestParameters().get(RESOURCE)))) {
+      final String audience = authentication.getOAuth2Request()
+        .getRefreshTokenRequest()
+        .getRequestParameters()
+        .get(RESOURCE);
```
giacomini, line 140: This was already computed in the `if`. Can we reuse that value?

iam-login-service/src/main/java/it/infn/mw/iam/core/error/InvalidResourceError.java
federicaagostini: Ok, maybe we can add a PR which updates all the licenses, otherwise the workflow would fail here https://github.com/indigo-iam/iam/blob/master/.github/workflows/maven.yml#L27.

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/IamOAuth2RequestFactory.java
federicaagostini: If the "resource" is not a valid URI, the whole request fails (see tests https://github.com/indigo-iam/iam/blob/issue-381/iam-login-service/src/test/java/it/infn/mw/iam/test/oauth/ResourceIndicatorTests.java).
At least this is my interpretation of the specification, which states

```
If the authorization server fails to parse the provided value(s) or does not consider the resource(s) acceptable, it should reject the request with an error response using the error code invalid_target as the value of the error parameter and can provide additional information regarding the reasons for the error using the error_description.
```

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/IamOAuth2RequestFactory.java
federicaagostini: No, just added the check on query parameters + fragments, and tests.

federicaagostini:
Superseded by https://github.com/indigo-iam/iam/pull/916.

PR #894, opened 2024-12-12
Exclude IAM optional groups from VOMS AC

rmiccoli:
In the context of WLCG profile, this PR removes from VOMS attribute certificate the groups those have "wlcg.optional-group" label set.

### DIFF
--- a/iam-voms-aa/src/test/java/it/infn/mw/voms/VomsAcTests.java
+++ b/iam-voms-aa/src/test/java/it/infn/mw/voms/VomsAcTests.java
@@ -269,6 +269,33 @@ public class VomsAcTests extends TestSupport {
     assertThat(attrs.getNotAfter(), lessThanOrEqualTo(Date.from(NOW_PLUS_12_HOURS)));
   }
 
+  @Test
+  public void optionalGroupIsReturnedForUserIfRequested() throws Exception {
+    IamAccount testAccount = setupTestUser();
+    IamGroup rootGroup = createVomsRootGroup();
+    IamGroup subGroup = createChildGroup(rootGroup, "sub");
+    IamGroup optionalGroup = createOptionalGroup(subGroup, "optional");
+
+    addAccountToGroup(testAccount, rootGroup);
+    addAccountToGroup(testAccount, subGroup);
+    addAccountToGroup(testAccount, optionalGroup);
+
+    byte[] xmlResponse = mvc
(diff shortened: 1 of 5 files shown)


enricovianello (approved):

iam-voms-aa/src/test/java/it/infn/mw/voms/VomsAcTests.java
```diff
@@ -243,6 +243,32 @@ public void allGroupsAreReturnedForUser() throws Exception {
...
+
+    VOMSResponse response = parser.parse(new ByteArrayInputStream(xmlResponse));
+    assertThat(response.hasErrors(), is(false));
+    VOMSAttribute attrs = getAttributeCertificate(response);
+    assertThat(attrs.getFQANs(), hasSize(2));
+    assertThat(attrs.getFQANs(), hasItem("/test"));
+    assertThat(attrs.getFQANs(), hasItem("/test/sub"));
+    assertThat(attrs.getNotAfter(), lessThanOrEqualTo(Date.from(NOW_PLUS_12_HOURS)));
+  }
+
```
giacomini, line 272: And what about a situation of a role in an optional group? e.g. what happens (or should happen) if you ask the role but don't ask for the optional group?

rmiccoli: For now, if you have, for instance
* root group `/test`
* optional group `/optional` (subgroup of the root group)
* role `VO-Admin` in the optional group
asking for `/test/optional/Role=VO-Admin` returns two fqans, `/test` and `/test/optional/Role=VO-Admin`.

giacomini: I wonder if it should return also the fqan `/test/optional`. It's strange to have a role in a group but not that group.

rmiccoli: Makes sense, but the same applies to tokens. If we have a subgroup called `sub` within an optional group `root`, when we request only `/root/sub`, we get `/root/sub` but not `/root`.

giacomini: Optional groups confirm to be a source of confusion and ambiguity. Leave it as is for the moment.

enricovianello (approved):

PR #898, opened 2024-12-17
Fix #809 JDBC session on same db

### LINKED ISSUE #809: JDBC for Cacheing/session storage

DonaldChung-HK:
Fix #809 by adding the JDBC session package. This is the spring default implementation, which uses additional tables on a db to store session data.

Replace https://github.com/indigo-iam/iam/pull/840 as I am unable to identify why the new data source object refuses to limit the amount of connection spawned, will work on it in the future.

Passed all tests on a `mariadb:10.11` container as flyway community don't support older db on my test host.

**Instruction to use JDBC as session storage**
the sample configuration in `application-jdbc-session.yml` details how to activate the flag, how often to clean up the session and the table name etc. 

If your `table-name` is set to `SPRING_SESSION` you will need to make the following 2 tables named `SPRING_SESSION` and `SPRING_SESSION_ATTRIBUTES` manually in your IAM database with the following SQL for example after the database is initialized by flyway.

```
CREATE TABLE IF NOT EXISTS SPRING_SESSION
(
    PRIMARY_ID            CHAR(36) NOT NULL,
    SESSION_ID            CHAR(36) NOT NULL,
    CREATION_TIME         BIGINT   NOT NULL,
    LAST_ACCESS_TIME      BIGINT   NOT NULL,
    MAX_INACTIVE_INTERVAL INT      NOT NULL,
    EXPIRY_TIME           BIGINT   NOT NULL,
    PRINCIPAL_NAME        VARCHAR(100),
    CONSTRAINT SPRING_SESSION_PK PRIMARY KEY (PRIMARY_ID)
) ENGINE=InnoDB ROW_FORMAT=DYNAMIC;

CREATE UNIQUE INDEX SPRING_SESSION_IX1 ON SPRING_SESSION (SESSION_ID);
CREATE INDEX SPRING_SESSION_IX2 ON SPRING_SESSION (EXPIRY_TIME);
CREATE INDEX SPRING_SESSION_IX3 ON SPRING_SESSION (PRINCIPAL_NAME);

CREATE TABLE IF NOT EXISTS SPRING_SESSION_ATTRIBUTES
(
    SESSION_PRIMARY_ID CHAR(36)     NOT NULL,
    ATTRIBUTE_NAME     VARCHAR(200) NOT NULL,
    ATTRIBUTE_BYTES    BLOB         NOT NULL,
    CONSTRAINT SPRING_SESSION_ATTRIBUTES_PK PRIMARY KEY (SESSION_PRIMARY_ID, ATTRIBUTE_NAME),
    CONSTRAINT SPRING_SESSION_ATTRIBUTES_FK FOREIGN KEY (SESSION_PRIMARY_ID) REFERENCES SPRING_SESSION (PRIMARY_ID) ON DELETE CASCADE
) ENGINE=InnoDB ROW_FORMAT=DYNAMIC;

```

You should see the session appearing in those databases as a result.

Thanks


federicaagostini:
Hi Donald, thank you for your PR.
We would like to not force the operator to manual create the needed tables, so you can write a migration for mysql such as [V105__add_aup_reminders.sql](https://github.com/indigo-iam/iam/blob/master/iam-persistence/src/main/resources/db/migration/mysql/V105__add_aup_reminders.sql) (using the prefix V106) and one for testing in h2 such as [V105__add_aup_reminders.sql](https://github.com/indigo-iam/iam/blob/master/iam-persistence/src/main/resources/db/migration/h2/V105__add_aup_reminders.sql). In this way, the tables will be created even if the profile is not active, but this is not a problem.

Also, if you could add few tests (for instance, the ones you have performed manually) it would be great. I guess you can include them in the [ExternalizedSessionDeviceCodeTests](https://github.com/indigo-iam/iam/blob/master/iam-login-service/src/test/java/it/infn/mw/iam/test/oauth/devicecode/ExternalizedSessionDeviceCodeTests.java) class.

DonaldChung-HK:
Hi @federicaagostini,

Thanks for your comment. I have added the SQL to create the table in both `h2` and `mysql`. However, I don't think additional tests are needed as this is just the default spring implementation so the upstream spring dev should have confirmed the functionality.

DonaldChung-HK:
Added test to ensure spring session JDBC is working properly

enricovianello (approved):

PR #902, opened 2025-01-09
Show amr claim in ID token

rmiccoli:
Reference: [RFC8176](https://datatracker.ietf.org/doc/html/rfc8176)

* When MFA is enabled, the value of `amr` claim in ID token is `["otp", "pwd"]`.
* When MFA is not enabled, the `amr` value is `["pwd"]`.


rmiccoli:
Superseded by https://github.com/indigo-iam/iam/pull/918/commits/2120a102e8c698cf88d06bdcd2a56cd7e24832e9

PR #916, opened 2025-02-04
Add support for resource parameter

federicaagostini:
Following [RFC8707](https://datatracker.ietf.org/doc/html/rfc8707), during an OAuth request the client may indicate a _resource_ parameter, which is the target service to which access is being requested (MUST be an absolute URI). The value of the resource parameter is returned as _aud_ claim in the JWT. Moreover, as the RFC states, the resources appearing in the JWT are a subset of the set of resources originally granted.

In case both `audience` and `resource` are provided, only the resource parameter will be included in the token (as _aud_ claim).

The current behavior of resource/audience parameter is descibed in the next table

| authorize endpoint | token endpoint | audience claim (IAM) | audience claim (WLCG) |
|------|------|------|------|
| resource = `https://storm.org` | resource (or aud) = `https://storm.org` | `https://storm.org` | `https://storm.org` |
| resource = `https://storm.org` | resource (or aud) = `https://dcache.org` | 400 Invalid Resource | 400 Invalid Resource |
| resource = `https://storm.org` | resource (or aud) = `https://storm.org https://dcache.org` | `https://storm.org` | `https://storm.org` |
| resource = `https://storm.org` | no resource | `https://storm.org` | `https://storm.org` |
| no resource | resource = `https://storm.org` | 400 Invalid Resource | 400 Invalid Resource |
| no resource | aud = `https://storm.org` | `https://storm.org` | `https://storm.org` |
| no resource | no aud, no resource | not present | `https://wlcg.cern.ch/jwt/v1/any` |


iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/granters/IamAuthorizationCodeTokenGranter.java
```diff
@@ -0,0 +1,106 @@
...
+
+    String pendingClientId = pendingOAuth2Request.getClientId();
+    String clientId = tokenRequest.getClientId();
+    if (clientId != null && !clientId.equals(pendingClientId)) {
+      throw new InvalidClientException("Client ID mismatch");
+    }
+
+    Map<String, String> combinedParameters =
+        new HashMap<String, String>(pendingOAuth2Request.getRequestParameters());
+    combinedParameters.putAll(parameters);
```
federicaagostini, line 91: Not sure why we have to include the authorization request parameters instead of the token request ones at the token creation, but it was the default behavior in Spring.

enricovianello:
We need to add the list of asked resources into the consent page

federicaagostini:
@giacomini what do you think about the above table?

federicaagostini:
Return the allowed resource as token response, in case a filter is applied (by the server) -- at least, apply same logic as for scopes.

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/wlcg/WLCGProfileAccessTokenBuilder.java
```diff
@@ -119,9 +121,29 @@ private void addAuthnInfoClaims(Builder builder, Set<String> scopes, UserInfo us
   }
 
   private void addAudience(Builder builder, OAuth2Authentication authentication) {
-    if (!hasAudienceRequest(authentication) && !hasRefreshTokenAudienceRequest(authentication)) {
+    if (!hasAnyAudience(authentication)) {
       builder.audience(ALL_AUDIENCES_VALUE);
```
enricovianello, line 125: we could check if builder.getClaims("aud") has been already defined. In case not, we must set it as the "any" value

enricovianello (changes requested):

enricovianello (approved):

rmiccoli (approved):

PR #918, opened 2025-02-05
Integrate MFA with external OIDC and SAML providers

rmiccoli:
## What has been implemented
When the user authenticates with an external OIDC/SAML provider
* via MFA, IAM skips the second-factor authentication 
* IAM then performs the second-factor authentication (if MFA is enabled)
  * in this case, the  `acr` claim with value (hardcoded) `https://refeds.org/profile/MFA` appears in ID token, access token and introspection response.

The ID token also shows the `amr` claim with value
* `["otp", "pwd"]` or `["otp", "oidc"]`, if MFA is enabled and user authenticates with username/password or via an external OIDC provider
* `["pwd"]` or `["oidc"]`, if MFA is NOT enabled and user only authenticates with username/password or via an external OIDC provider

> linked to https://github.com/indigo-iam/OpenID-Connect-Java-Spring-Server/pull/26

### Notes
The `oidc` and `saml` values are not included in the [RFC8176](https://www.rfc-editor.org/rfc/rfc8176.html), so they should be treated as an extension and documented.

### References
* [RFC9470](https://www.rfc-editor.org/rfc/rfc9470.html)
* [REFEDS MFA profile](https://refeds.org/profile/mfa)
* [RFC8176](https://www.rfc-editor.org/rfc/rfc8176.html)


iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/common/BaseAccessTokenBuilder.java
```diff
@@ -170,4 +182,20 @@ protected JWTClaimsSet.Builder baseJWTSetup(OAuth2AccessTokenEntity token,
...
+    Optional<IamAccount> account = accountUtils.getAuthenticatedUserAccount(authentication);
+    if (authentication.getUserAuthentication() != null
+        && authentication.getUserAuthentication().getDetails() instanceof Map<?, ?> details
+        && details.get("acr") != null) {
+      builder.claim("acr", details.get("acr"));
+    } else {
+      if (account.isPresent()) {
+        Optional<IamTotpMfa> totpMfaOptional = totpMfaRepository.findByAccount(account.get());
+        if (totpMfaOptional.isPresent() && totpMfaOptional.get().isActive()) {
+          builder.claim("acr", "https://referds.org/profile/MFA");
```
enricovianello, line 196: Could we have a case when we don't want to stress users to do the 2FA but we require this from the remote SSO? And we want to fail in case is not applied? Probably this will be required by someone. What about having the "default acr value to be required" as a configuration parameter for the remote OIDC provider?

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/common/BaseIdTokenCustomizer.java
```diff
@@ -67,4 +78,30 @@ protected final void includeLabelsInIdToken(Builder idClaims, IamAccount account
...
 
+  protected final void includeAmrAndAcrClaimsIfNeeded(OAuth2Request request, Builder builder,
+      OAuth2AccessTokenEntity accessToken) {
+    Object amrClaim = request.getExtensions().get("amr");
+
+    if (amrClaim instanceof String amrString) {
+      try {
+        ObjectMapper objectMapper = new ObjectMapper();
+        List<String> amrList =
+            objectMapper.readValue(amrString, new TypeReference<List<String>>() {});
```
enricovianello, line 89: so the amrString is a JSON object?

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/common/BaseIntrospectionHelper.java
```diff
@@ -108,6 +108,20 @@ protected void addAudience(Map<String, Object> introspectionResult,
...
+    try {
+      Object acr = accessToken.getJwt().getJWTClaimsSet().getClaim("acr");
+      if (acr instanceof String acrString) {
+        introspectionResult.put("acr", acrString);
+      }
+    } catch (ParseException e) {
+      LOG.error("Error getting acr claim out of access token: {}", e.getMessage(), e);
+    }
+
+  }
```
enricovianello, line 123: We're moving in a direction that the content of the access token will be printed claim by claim into the introspection endpoint but it something on me on my branch as soon as this PR will be merged and my branch rebased

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/iam/IamJWTProfileIdTokenCustomizer.java
```diff
@@ -59,7 +59,10 @@ public void customizeIdTokenClaims(Builder idClaims, ClientDetailsEntity client,
       .filter(ADDITIONAL_CLAIMS::contains)
       .forEach(c -> idClaims.claim(c, claimValueHelper.getClaimValueFromUserInfo(c, info)));
 
+    includeAmrAndAcrClaimsIfNeeded(request, idClaims, accessToken);
+
     includeLabelsInIdToken(idClaims, account);
-  }
 
+    idClaims.claim("voperson_id", account.getUserInfo().getSub());
```
enricovianello, line 66: Have you rebased on develop?

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/keycloak/KeycloakProfileAccessTokenBuilder.java
```diff
@@ -67,6 +71,8 @@ public JWTClaimsSet buildAccessToken(OAuth2AccessTokenEntity token,
       }
     }
 
+    addAcrClaimIfNeeded(builder, authentication);
```
enricovianello, line 74: if all the profiles are calling this logic, we could simply include it inside the parent constructor, or not?

iam-login-service/src/main/java/it/infn/mw/iam/authn/oidc/OidcExternalAuthenticationToken.java
```diff
@@ -52,14 +53,55 @@ public OidcExternalAuthenticationToken(OIDCAuthenticationToken authn, Date token
     super(authn, tokenExpiration, principal, credentials, authorities);
   }
 
+  @Override
+  public boolean equals(Object obj) {
```
enricovianello, line 57: Why we need to explicit this equals (+ hashCode)? I cannot find where it's used.

enricovianello (commented):
I started this review with some questions

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/common/BaseIdTokenCustomizer.java
rmiccoli: > so the amrString is a JSON object?

Yes, it's a JSON array of strings

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/common/BaseAccessTokenBuilder.java
rmiccoli: We have already discussed this verbally, here we are just building the access token by inserting the `acr` claim

enricovianello (approved):

PR #919, opened 2025-02-06
Update default VOMS AA application.yml

jouvin:
This PR replaces the default VOMS AA `application.yml` by a version using environment variables to configure it, to avoid a site the need to deploy its own specific version. The variables used are those already used by the login service plus a few ones specific to the VOMS AA service:

- `VOMS_AA_HOST`: the host where VOMS AA service is running (no default)
- `VOMS_AA_PORT`: the port the VOMS AA service binds to (no default)
- `VOMS_AA_VO`: the VO served by the VOMS AA service (no default)
- `VOMS_AA_USE_LEGACY_FQAN_ENCODING`: defaults to `true`

This change should not have any impact on existing VOMS AA instances as the current default `application.yml` is not usable, meaning that sites will have deployed their own specific version that will continue to override the default one.

- [ ] Once/if this PR is accepted, the `application.yml` contents to use (similar to this one) should be removed from the [VOMS AA documentation](https://github.com/indigo-iam/iam-website/blob/v1.11.0/content/en/docs/tasks/deployment/voms/_index.md).

### DIFF
--- a/iam-voms-aa/src/main/resources/application.yml
+++ b/iam-voms-aa/src/main/resources/application.yml
@@ -1,46 +1,41 @@
-#
-# Copyright (c) Istituto Nazionale di Fisica Nucleare (INFN). 2016-2021
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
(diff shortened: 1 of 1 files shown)


enricovianello:
Superseded by https://github.com/indigo-iam/iam/pull/1178 
Included into next IAM v1.14.0

PR #925, opened 2025-02-26
Fix the subgroup name update during PUT to scim/Groups/{id}

### LINKED ISSUE #924: PUT to /scim/Groups/{id} does not update the subgroup name correctly
If we want to change the subgroup name A/B to A/BB, we must make a PUT call to the `scim/Groups/{id}` endpoint.
The displayName cannot contain the character `\`, so we will enter BB and then it should be updated and shown in the dashboard as A/BB.

Currently, the dashboard only shows BB (unlike POST to `scim/Groups` when a subgroup is created).

rmiccoli:
The subgroup name is now updated correctly according to this rule:
* `parentGroup/subgroup`.

In addition, when a root group or subgroup is updated, all child groups, if any, are also updated.


federicaagostini:
> The subgroup name is now updated correctly according to this rule:
> 
>     * `parentGroup/subgroup`.
> 
> 
> In addition, when a root group or subgroup is updated, all child groups, if any, are also updated.

We have decided to change the logic such to allow updating only the last child group and do not allow cascade update because of inconsistency issues.

Moreover, in this PR we allow also to update the group description.

The group labels should also be updated, but not included in this PR: anyway there is an ignored test case called `testUpdateGroupLabelsSuccessResponse` which can be enabled in future for this specific use-case.

enricovianello (approved):

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimGroupProvisioning.java
```diff
@@ -250,19 +250,43 @@ private Supplier<ScimResourceNotFoundException> noGroupMappedToId(String id) {
...
+      String fullName =
+          String.format("%s/%s", groupToUpdate.getParentGroup().getName(), displayName);
+      fullNameSanityChecks(fullName);
+
+      groupToUpdate.setName(fullName);
+    } else {
+      groupToUpdate.setName(displayName);
+    }
+
+    if (!groupToUpdate.getChildrenGroups().isEmpty()) {
```
enricovianello: i'd move this check before the parent one.

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimGroupProvisioning.java
```diff
@@ -250,19 +250,43 @@ private Supplier<ScimResourceNotFoundException> noGroupMappedToId(String id) {
...
-    groupService.updateGroup(oldGroup, newGroup);
+    IamGroup groupToUpdate = groupService.findByUuid(id).orElseThrow(noGroupMappedToId(id));
+    String oldGroupName = groupToUpdate.getName();
+
+    if (groupToUpdate.getParentGroup() != null) {
+      String parentGroupUuid = groupToUpdate.getParentGroup().getUuid();
+
+      groupService.findByUuid(parentGroupUuid)
+        .orElseThrow(() -> new ScimResourceNotFoundException(
+            String.format("Parent group '%s' not found", parentGroupUuid)));
```
enricovianello: Not necessary, it's something ensured by JPA. If it's not null, it means it exists. You can directly access `groupToUpdate.getParentGroup().getName()` if `groupToUpdate.getParentGroup()` is not null

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimGroupProvisioning.java
```diff
@@ -250,19 +250,43 @@ private Supplier<ScimResourceNotFoundException> noGroupMappedToId(String id) {
   @Override
   public ScimGroup replace(String id, ScimGroup scimItemToBeReplaced) {
 
-    IamGroup oldGroup = groupService.findByUuid(id).orElseThrow(noGroupMappedToId(id));
-
     String displayName = scimItemToBeReplaced.getDisplayName();
```
enricovianello: I'd rename also scimItemToBeReplaced, but for sure I'd use newDisplayName for displayName. Makes things clearer later

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimGroupProvisioning.java
```diff
@@ -250,19 +250,43 @@ private Supplier<ScimResourceNotFoundException> noGroupMappedToId(String id) {
...
     displayNameSanityChecks(displayName);
 
     if (!isGroupNameAvailable(displayName, id)) {
       throw new ScimResourceExistsException(displayName + " is already mapped to another group");
     }
 
-    IamGroup newGroup = converter.entityFromDto(scimItemToBeReplaced);
-    groupService.updateGroup(oldGroup, newGroup);
+    IamGroup groupToUpdate = groupService.findByUuid(id).orElseThrow(noGroupMappedToId(id));
```
enricovianello: I'd move this again as the first operation inside this method. Before checking the received displayName we need to check if the id is valid.

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimGroupProvisioning.java
```diff
@@ -250,19 +250,43 @@ private Supplier<ScimResourceNotFoundException> noGroupMappedToId(String id) {
...
+
+    if (!groupToUpdate.getChildrenGroups().isEmpty()) {
+      throw new IllegalArgumentException("The current group contains child group(s)");
+    }
+
+    if (scimItemToBeReplaced.getIndigoGroup().getDescription() != null) {
+      groupToUpdate.setDescription(scimItemToBeReplaced.getIndigoGroup().getDescription());
+    }
+
+    groupService.updateGroup(oldGroupName, groupToUpdate);
```
enricovianello: I'd expect id to be used here and not the oldGroupName.. strange

iam-login-service/src/main/java/it/infn/mw/iam/core/group/DefaultIamGroupService.java
```diff
@@ -202,19 +202,11 @@ public void setApplicationEventPublisher(ApplicationEventPublisher applicationEv
...
-  public IamGroup updateGroup(IamGroup oldGroup, IamGroup newGroup) {
-
-    newGroup.setId(oldGroup.getId());
-    newGroup.setUuid(oldGroup.getUuid());
-    newGroup.setCreationTime(oldGroup.getCreationTime());
-
-    newGroup.setDescription(oldGroup.getDescription());
-    newGroup.setParentGroup(oldGroup.getParentGroup());
-    newGroup.setChildrenGroups(oldGroup.getChildrenGroups());
+  public IamGroup updateGroup(String oldGroupName, IamGroup newGroup) {
```
enricovianello, line 205: Using GroupId should be used here

enricovianello: And all the logic used inside ScimGroupProvisioning should be moved here.
e.g.
- check id existance
- verify children
- verify parent
- update display name or description
... etc

It's the service that should do the work.

enricovianello (changes requested):
Mainly the logic should be moved inside the group service. Other minor fixes.

PR #926, opened 2025-02-26
Integrate mfa with x.509

garaimanoj:


iam-login-service/src/main/java/it/infn/mw/iam/authn/multi_factor_authentication/IamAuthenticationMethodReference.java
```diff
@@ -27,7 +27,7 @@ public enum AuthenticationMethodReferenceValues {
     // Consult here for standardised reference values -
     // https://datatracker.ietf.org/doc/html/rfc8176
```
enricovianello, line 28: Reading this RFC, we should probably not use "x509" (because it doesn't exist) and replace this string with "swk" which means "Proof-of-Possession (PoP) of a software-secured key". In fact, it refers to RFC4211 https://datatracker.ietf.org/doc/html/rfc4211

iam-login-service/src/main/java/it/infn/mw/iam/authn/multi_factor_authentication/IamAuthenticationMethodReference.java
```diff
@@ -27,7 +27,7 @@ public enum AuthenticationMethodReferenceValues {
...
 
-    PASSWORD("pwd"), ONE_TIME_PASSWORD("otp");
+    PASSWORD("pwd"), ONE_TIME_PASSWORD("otp"), X509("x509");
```
enricovianello, line 30: so, it should be `X509("swk")`

iam-login-service/src/main/java/it/infn/mw/iam/authn/multi_factor_authentication/MfaVerifyController.java
```diff
@@ -59,10 +66,27 @@ public String getVerifyMfaView(Authentication authentication, ModelMap model) {
...
 
+  private void setAuthentication(PreAuthenticatedAuthenticationToken preAuthenticatedAuthenticationToken) {
+    Set<GrantedAuthority> authenticatedAuthorities = new HashSet<>(
+        preAuthenticatedAuthenticationToken.getAuthorities());
+    if (preAuthenticatedAuthenticationToken.getPrincipal() instanceof User user) {
+      ExtendedAuthenticationToken token = new ExtendedAuthenticationToken(user.getUsername(), "SECRET",
+          authenticatedAuthorities);
+      token.setAuthenticated(false);
+      token.setAuthenticationMethodReferences(Set.of(
+          new IamAuthenticationMethodReference(X509.getValue())));
```
enricovianello, line 84: why not Set.of(IamAuthenticationMethodReference.X509)?

enricovianello (changes requested):
Small adjustments :)

iam-login-service/src/main/java/it/infn/mw/iam/authn/multi_factor_authentication/MfaVerifyController.java
garaimanoj: With that change getting error `X509 cannot be resolved or is not a field`

enricovianello: You're right :)

iam-login-service/src/main/java/it/infn/mw/iam/authn/multi_factor_authentication/IamAuthenticationMethodReference.java
garaimanoj: Done

rmiccoli (approved):

enricovianello (approved):

PR #928, opened 2025-02-27
Implement JIT account provisioning for OPs

### LINKED ISSUE #927: Configure JIT account provisioning for trusted OPs

### LINKED ISSUE #1011: Delete cleanup task for auto-provisioned accounts
For SAML providers, there is a cleanup task, under the jit account provisioning configuration, that delete inactive accounts.
By default it is disabled, but, if enabled, the auto-provisioned accounts whose last login is less than a certain time (default is 15 days) are deleted.
By introducing the functionality of providing jit accounts also for OIDC providers, the cleanup activity will not differentiate between accounts registered via OIDC providers or with SAML providers.
At the moment, since unused, we delete this cleanup.

rmiccoli:
In order to enable JIT account provisioning, you need to set
* `IAM_OIDC_JIT_ACCOUNT_PROVISIONING_ENABLED=true`
* `IAM_OIDC_JIT_ACCOUNT_PROVISIONING_TRUSTED_IDPS=all` or equal to a specific trusted OIDC issuer (e.g. `https://accounts.google.com`).

Just note that, if JIT account provisioning is enabled and a user tries to register with an _untrusted_ issuer, they get an error. This is true also for SAML.


iam-login-service/src/main/java/it/infn/mw/iam/config/oidc/OidcConfiguration.java
```diff
@@ -232,4 +251,9 @@ UserInfoFetcher userInfoFetcher() {
   OidcTokenRequestor tokenRequestor(RestTemplateFactory restTemplateFactory, ObjectMapper mapper) {
     return new DefaultOidcTokenRequestor(restTemplateFactory, mapper);
   }
+
+  @Override
+  public void configureTasks(ScheduledTaskRegistrar taskRegistrar) {
+    cleanupScheduler.scheduleCleanupTask(taskRegistrar, jitProperties, accountService, "OIDC");
```
enricovianello: Both this Configuration class and the SAML one schedule in practice the same cleanup task which deletes users independently from the fact they have been created through SAML or OIDC. Probably we should move a unique task creation/configuration inside TaskConfig class

enricovianello (dismissed):
I think we are scheduling two identical tasks (in case both oidc and saml profiles are used)
Probably we can add a Task defined somewhere else with @Profile({"oidc", "saml"})

iam-login-service/src/main/java/it/infn/mw/iam/config/oidc/OidcConfiguration.java
rmiccoli: We decided to delete the cleanup task for now, as it is unused.

enricovianello (commented):
I think that we can introduce this as "experimental" and get feedback from users.
Having added some lines of code I cannot approve it, we need another reviewer :)

enricovianello (approved):

PR #937, opened 2025-03-14
Support DER & URL-Encoded PEM client certificates

balciiberk:
During client certificate authentication:
- Supports new nginx variable `$ssl_client_escaped_cert` to be used in the header. This does URL-encoding to the PEM certificates.
- Supports HAProxy by
  - supporting base64 encoded DER certificates in the header.
  - accepting `X509_V_OK` at the `X-SSL-Client-Verify` header
- Keeps backwards compatibility by keeping support of the old nginx variable `$ssl_client_cert`

After this update, HAproxy can also be used for client certificate authentication by setting the following headers:
```
    http-request set-header X-SSL-Client-Cert %{+Q}[ssl_c_der,base64]
    http-request set-header X-SSL-Client-S-Dn %{+Q}[ssl_c_s_dn]
    http-request set-header X-SSL-Client-I-Dn %{+Q}[ssl_c_i_dn]
    http-request set-header X-SSL-Client-Serial %{+Q}[ssl_c_serial]
    http-request set-header X-SSL-Client-V-Start %{+Q}[ssl_c_notbefore]
    http-request set-header X-SSL-Client-V-End %{+Q}[ssl_c_notafter]
    http-request set-header X-SSL-Protocol %[ssl_fc_protocol]
    http-request set-header X-SSL-Client-Verify %{+Q}[ssl_c_verify,x509_v_err_str]
```

I changed the name of the `PEMX509CertificateChainParser.java` to `X509CertificateChainParserImpl.java` because after these changes, it's not only parsing PEM. It's parsing both PEM and DER. If this makes the Git history messy, I can explore another way to do it.


giacomini:
About the filename change, did you use `git mv`?

balciiberk:
I didn't because I didn't know this existed. Now I see that, for it to keep the history I need to use `git mv` and change the file in a separate commit. I tried reverting and doing that but it didn't work. I will delete the branch and redo it using git mv.

PR #938, opened 2025-03-14
Support DER & URL-Encoded PEM client certificates

balciiberk:
During client certificate authentication:
- Supports new nginx variable `$ssl_client_escaped_cert` to be used in the header. This does URL-encoding to the PEM certificates.
- Supports HAProxy by
  - supporting base64 encoded DER certificates in the header.
  - accepting `X509_V_OK` at the `X-SSL-Client-Verify` header
- Keeps backwards compatibility by keeping support of the old nginx variable `$ssl_client_cert`

After this update, HAproxy can also be used for client certificate authentication by setting the following headers:
```
    http-request set-header X-SSL-Client-Cert %{+Q}[ssl_c_der,base64]
    http-request set-header X-SSL-Client-S-Dn %{+Q}[ssl_c_s_dn]
    http-request set-header X-SSL-Client-I-Dn %{+Q}[ssl_c_i_dn]
    http-request set-header X-SSL-Client-Serial %{+Q}[ssl_c_serial]
    http-request set-header X-SSL-Client-V-Start %{+Q}[ssl_c_notbefore]
    http-request set-header X-SSL-Client-V-End %{+Q}[ssl_c_notafter]
    http-request set-header X-SSL-Protocol %[ssl_fc_protocol]
    http-request set-header X-SSL-Client-Verify %{+Q}[ssl_c_verify,x509_v_err_str]
```

I changed the name of the `PEMX509CertificateChainParser.java` to `X509CertificateChainParserImpl.java` because after these changes, it's not only parsing PEM. It's parsing both PEM and DER. If this makes the Git history messy, I can explore another way to do it.


iam-login-service/src/main/java/it/infn/mw/iam/authn/x509/X509CertificateChainParserImpl.java
```diff
@@ -0,0 +1,105 @@
...
+
+  private boolean isPEMFormat(String certString) {
+    return certString.contains("-----BEGIN");
+  }
+
+
+  private X509CertificateChainParsingResult parsePEM(String certString) {
+    String pemString;
+
+    if (certString.contains("\t")) {
```
enricovianello, line 60: Just for my knowledge, why these checks have been introduced? Did you follow some kind of ref. / rfc? Thanks :)
For all the other fixes/changes it's ok for me this PR.

enricovianello (commented):
I have only one question, you can find it near the involved lines of code

iam-login-service/src/main/java/it/infn/mw/iam/authn/x509/X509CertificateChainParserImpl.java
balciiberk: Because the deprecated nginx variable `$ssl_client_cert` exchanges all `\n` to `\t`. That's why we were already doing
`pemString = certString.replace("\t", "\n");`

But the `$ssl_client_escaped_cert` (which is what the nginx document suggests to use instead), uses URL encoding, which encodes `\n` as `%0A`.

I found these in the [nginx docs](https://nginx.org/en/docs/http/ngx_http_ssl_module.html)

giacomini:
It looks ok to me. I have one question: how did you generate the .der file used for the tests?

balciiberk:
> It looks ok to me. I have one question: how did you generate the .der file used for the tests?

I just converted the already existing test0.cert.pem in the same directory with a command like:

`openssl x509 -in test0.cert.pem -outform der -out test0.cert.der`

enricovianello (approved):

PR #944, opened 2025-03-24
Delete useless forbidden calls from current dashbaord

rmiccoli:
Group labels are now retrieved from the `/iam/group/{groupId}/labels` endpoint, accessible to all authenticated users.

### DIFF
--- a/iam-login-service/src/main/java/it/infn/mw/iam/api/group/GroupLabelsController.java
+++ b/iam-login-service/src/main/java/it/infn/mw/iam/api/group/GroupLabelsController.java
@@ -67,7 +67,7 @@ public class GroupLabelsController {
   }
 
   @GetMapping
-  @PreAuthorize("#iam.hasScope('iam:admin.read') or #iam.hasDashboardRole('ROLE_ADMIN') or hasRole('READER') or #iam.isGroupManager(#id)")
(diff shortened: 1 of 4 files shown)


enricovianello (changes requested):
I'd change this AuthZ by opening it to all the authenticated users (`hasRole('ROLE_USER')`) https://github.com/indigo-iam/iam/blob/master/iam-login-service/src/main/java/it/infn/mw/iam/api/group/GroupLabelsController.java#L70 and then I'd change the self.groupLabels implementation with a call to this updated endpoint instead of the SCIM one. I think the labels associated to a group must be visible to all the users. There's no security problems on knowing it's a voms role o an optional group I think.

rmiccoli:
> I'd change this AuthZ by opening it to all the authenticated users (`hasRole('ROLE_USER')`) https://github.com/indigo-iam/iam/blob/master/iam-login-service/src/main/java/it/infn/mw/iam/api/group/GroupLabelsController.java#L70 and then I'd change the self.groupLabels implementation with a call to this updated endpoint instead of the SCIM one. I think the labels associated to a group must be visible to all the users. There's no security problems on knowing it's a voms role o an optional group I think.

Done! :)

enricovianello (approved):

PR #946, opened 2025-03-26
Implement SCIM filtering

### LINKED ISSUE #931: GET /scim/Users should allow filtering by fields
Currently the GET /scim/Users endpoint doesn't allow [SCIM filtering](https://datatracker.ietf.org/doc/html/rfc7644#section-3.4.2.2)

As requested in the INDIGO IAM Hackathon, it would be nice if it was possible to filter by fields in the SCIM API. Example usages would be, filtering by a username to get a user, or filtering by the active field to get only active users, etc.

If this is implemented, the [documentation](https://indigo-iam.github.io/v/v1.11.0/docs/reference/api/scim-api/#get-scimusers) needs to be updated as it currently says "SCIM Filtering and sorting of results are currently not supported."

S203932:
Allowing for filtering according to the following Operators: 
- eq
- co
And with the following attributes: 
- familyName
- givenName
- username
- active
- emails


iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
```diff
@@ -112,59 +112,56 @@ public ScimUserProvisioning(IamAccountService accountService,
...
         // Need to look at all instances that corresponds to the operator
-        while(-1 != filtersParameter.indexOf(operator.toString(), from)){
+        while (-1 != filtersParameter.indexOf(operator.toString(), from)) {
 
-           // Looking for the first instance of the given operator in the request
+          // Looking for the first instance of the given operator in the request
           int index = filtersParameter.indexOf(operator.toString(), from);
 
-          index = index==0?index=1:index;
+          index = index == 0 ? index = 1 : index;
```
giacomini, line 146: this I don't understand. does it mean that if index is 0, it should become 1? maybe an `if` is clearer. and in any case I don't think you need `index = ` before `1`.

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
```diff
@@ -174,12 +171,12 @@ private ArrayList<String> parseFilters(final String filtersParameter) {
...
 
 
-  private boolean filterEvaluation(ArrayList<String> parsedFilters){
-    if(parsedFilters.size()==3){
-      for(Attributes attribute : Attributes.values()){  
-        if(attribute.toString().equalsIgnoreCase(parsedFilters.get(0))){
-          for(AttributeOperators operator : AttributeOperators.values()){
-            if(operator.toString().equalsIgnoreCase(parsedFilters.get(1))){
+  private boolean filterEvaluation(ArrayList<String> parsedFilters) {
+    if (parsedFilters.size() == 3) {
```
giacomini, line 175: can we reduce the overall nesting? maybe with an early return if parsedFilters.size() is not 3?

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
```diff
@@ -112,59 +112,56 @@ public ScimUserProvisioning(IamAccountService accountService,
...
-            String parameter = filtersParameter.substring(0, index -1);
+            String parameter = filtersParameter.substring(0, index - 1);
 
             // The value is from the operator until the end
-            String value = filtersParameter.substring(index + operator.name().length() + 1, filtersParameter.length());
-            
+            String value = filtersParameter.substring(index + operator.name().length() + 1,
+                filtersParameter.length());
+
             result.add(parameter);
```
giacomini, line 159: This code is probably pre-existing, but I wonder if it would be better to represent a filter not as an array of three strings (which implies using get(i)), but as as a struct/class with three named members. Sorry for using a C++ terminology.

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
S203932: Yes, if index is equal to 0 than it should become 1. It is to make sure it does not get stuck by finding the next instance and keep moving it forward. 
Yes, one can omit the `index = `  before the `1`, haven't been using shorthand that much :sweat_smile:
I can do an `if` if that is preferred, just did the shorthand as it is shorter, line-wise

giacomini: In this case I think an `if` would be clearer
```suggestion
          if (index == 0) {
            index = 1;
          };
```

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
giacomini: I leave this to @enricovianello to decide if he wants to profit from this opportunity to restructure the code and how.

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
S203932: Sure thing, I commited your suggestion

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
S203932: > Personally, I prefer this approach as it keeps the code cleaner and more readable. You could create the Filter class (e.g., ScimFilter.class) under /iam/api/scim/model.

The Arraylist has now been replaced with the ScimFilter object from the class in  `/iam/api/scim/model/ScimFilter.java

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
```diff
@@ -112,59 +112,56 @@ public ScimUserProvisioning(IamAccountService accountService,
...
-           // Looking for the first instance of the given operator in the request
+          // Looking for the first instance of the given operator in the request
           int index = filtersParameter.indexOf(operator.toString(), from);
 
-          index = index==0?index=1:index;
+          index = index == 0 ? index = 1 : index;
 
           // If there is an occurence and it's surrounded by spaces
-          if(index > 0 && filtersParameter.charAt(index-1) == ' ' && filtersParameter.charAt(index + operator.name().length()) == ' '){
+          if (index > 0 && filtersParameter.charAt(index - 1) == ' '
```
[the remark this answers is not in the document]
S203932: @giacomini I have no removed the check. 

Yes `index` could be `-1` if there was no instance of the given `operator` in the filtering string. 
However, this is not possible given that the code does not enter the `while` - loop in line `140` if that is the case, i.e. it should not be possible for `index` to be `-1` and the given `if` statement.

enricovianello (approved):

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/model/ScimFilter.java
```diff
@@ -0,0 +1,47 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package it.infn.mw.iam.api.scim.model;
+
+public class ScimFilter {
+
+    private final String attribute;
```
enricovianello, line 21: We could use enum as types for both `attribute` and `operator`. The enum defined inside ScimUserProvisioning class

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
```diff
@@ -107,6 +111,163 @@ public ScimUserProvisioning(IamAccountService accountService,
...
+
+  // All the supported filter attributes
+  private enum Attributes {
+    // All related to the name of the user
+    familyName, givenName, username,
+
+    // Other info
+    active, emails
+  }
+
```
enricovianello, line 128: I always prefer having types in separated files. I'd move these enums outside this class. If missing, we can create a sub-package inside it.infn.mw.iam.api.scim.provisioning called model for example.

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
```diff
@@ -107,6 +111,163 @@ public ScimUserProvisioning(IamAccountService accountService,
...
+
+
+  private ScimFilter parseFilters(final String filtersParameter) {
+
```
enricovianello, line 132: My suggestion here is to create a ScimFilterParser that implements a method like:

```
public ScimFilter parse(String filter)
```

We can do this parsing with less if-then-else and substring calls in order to obtain a more readable codebase.
We can use regex to parse the expected string, for example:

```
  String regex = "([\\w\\.]+)\\s+(eq|co)\\s+\"([^\"]*)\"";
  Pattern pattern = Pattern.compile(regex);
  Matcher matcher = pattern.matcher(filtersParameter);

  if (!matcher.matches()) {
    throw invalidFilter(filtersParameter);
 }
```

Then you can use the matcher to get the String parts of the ScimFilter:

```
  String attributeStr = matcher.group(1);
  String operatorStr = matcher.group(2);
  String value = matcher.group(3);
```
and use the Enum native methods to get the correct Attribute/Operator and create the ScimFilter object:

```
  Attribute attribute = parseAttribute(attributeStr);
  AttributeOperator operator = AttributeOperator.valueOf(operatorStr);
  return new ScimFilter(attribute, operator, value);
```

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
```diff
@@ -107,6 +111,163 @@ public ScimUserProvisioning(IamAccountService accountService,
...
+            from = index + 1;
+          }
+        }
+      }
+    }
+    throw invalidFilter(filtersParameter);
+  }
+
+
+  private boolean filterEvaluation(ScimFilter parsedFilters) {
```
enricovianello, line 172: this method becomes useless if we use enums and is no more necessary to evaluate the ScimFilter object

enricovianello (changes requested):
I'm suggesting things that should make codebase more readable. It's the start of a review but I think I've suggested a good amount of things and then I'll have a second review as soon as these changes are in place. Sorry for this amount of changes :)

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimProvisioning.java
```diff
@@ -35,4 +35,7 @@ public interface ScimProvisioning<T, E> {
 
   ScimListResponse<T> list(ScimPageRequest params);
 
+  // The method below cannot exist until it is also implemented for ScimGroupProvisioning.java
+  // ScimListResponse<T> customList(ScimPageRequest params, String filter);
```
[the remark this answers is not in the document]
S203932: It has been done :+1:

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/controller/ScimUserController.java
```diff
@@ -79,25 +80,45 @@ private Set<String> parseAttributes(final String attributesParameter) {
...
       @RequestParam(required = false) final Integer startIndex,
-      @RequestParam(required = false) final String attributes) {
+      @RequestParam(required = false) final String attributes,
+      @RequestParam(required = false) final String filters) {
+
+
+    MappingJacksonValue wrapper;
+
+
+    if (filters != null) {
```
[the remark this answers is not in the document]
S203932: It has been done :+1: 

The method below no longer exists.
`public ScimListResponse<ScimUser> list(pr) {
 ...
}`

The `public ScimListResponse<ScimUser> list(pr)` has been integrated into `public ScimListResponse<ScimUser> list(pr, filters)`.

`list(pr)` is still in the interface (as it is used in `ScimGroupProvisioning`), but throws an exception in `ScimUserProvisioning` (which there is a test for).

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/controller/ScimControllerSupport.java
```diff
@@ -28,6 +28,13 @@ protected ScimPageRequest buildUserPageRequest(Integer count, Integer startIndex
     return buildPageRequest(count, startIndex, SCIM_USER_MAX_PAGE_SIZE);
   }
 
+
+  // To browse through all users, and not just the ones within the page limit
+  protected ScimPageRequest buildcustomUserPageRequest(Integer count, Integer startIndex,
+      Integer totalUsers) {
+    return buildPageRequest(count, startIndex, totalUsers);
+  }
```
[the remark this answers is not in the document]
S203932: Nope, it has been deleted :+1:

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/model/ScimFilter.java
S203932: It has been done :+1: 

`ScimFilter` now uses the enums `ScimFilterAttributes` and `ScimFilterOperators`

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
S203932: It has been done :+1: 
They have been moved to their own enum classes in the folder iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/model/`. 
The classes are `ScimFilterAttributes` and `ScimFilterOperators`

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
S203932: It has been removed :+1:

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimGroupProvisioning.java
```diff
@@ -233,6 +241,11 @@ public ScimListResponse<ScimGroup> list(ScimPageRequest pageRequest) {
     return builder.build();
   }
 
+  @Override
+  public ScimListResponse<ScimGroup> list(final ScimPageRequest params, String filter) {
+    throw methodNotSupported("list(final ScimPageRequest params, String filter)");
```
enricovianello: We could simplify the implementation of this by using the standard `UnsupportedOperationException`:

```
throw new UnsupportedOperationException("Unsupported filtered list");
```

Probably a simpler and better solution could be calling the `list` method defined before and ignoring the list parameter. We could add a log.warn message to track this wrong call on log.

enricovianello: This makes useless and removable the ScimInvalidMethod exception

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
```diff
@@ -107,6 +114,226 @@ public ScimUserProvisioning(IamAccountService accountService,
...
+    String attributeStr = matcher.group(1);
+    String operatorStr = matcher.group(2);
+    String value = matcher.group(3);
+
+
+    ScimFilterAttributes attribute =
+        ScimFilterAttributes.parseAttribute(attributeStr.toLowerCase());
+    ScimFilterOperators operator = ScimFilterOperators.parseOperator(operatorStr.toLowerCase());
+
+
```
enricovianello: too many empty lines

enricovianello:
Conflicts to be fixed

iam-persistence/src/main/java/it/infn/mw/iam/persistence/repository/IamAccountRepository.java
```diff
@@ -151,4 +154,62 @@ Page<IamAccount> findByLabelNameAndValue(@Param("name") String name, @Param("val
   @Query("delete from IamAccountGroupMembership")
   void deleteAllAccountGroupMemberships();
 
+
+  @Query("select a from IamAccount a join a.userInfo ui where ui.givenName = :givenName")
+  Optional<Page<IamAccount>> findByGivenName(@Param("givenName") String givenName,
```
enricovianello: Is really Optional necessary? Probably the returned Page when there's no results is a Page with size 0 and no contents.

enricovianello (changes requested):
Probably a useless Optional on repo is the main (little) change

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
S203932: They've been removed :+1:

iam-persistence/src/main/java/it/infn/mw/iam/persistence/repository/IamAccountRepository.java
S203932: You're correct, it was a bit of a barricade to make sure it would always run and not throw any errors (error handling is still in place, but a more reasonable amount). 
Optional has been removed and functions with the main use of these methods (in file `ScimUserProvisioning.java`) has been simplified.

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimGroupProvisioning.java
S203932: I've now changed it, so the method `list(final ScimPageRequest params, String filter)` gives a warning to the logger and then calls the `list(ScimPageRequest pageRequest)` (which doesn't have the `filter` parameter). 
The corresponding test has been changed as well to match this.

S203932: The method `list(final ScimPageRequest params, String filter)` now produces a warning in the logger and then calls the method `list(ScimPageRequest pageRequest)`, whilst ignoring the `filter` parameter. 
The corresponding test has been changed to reflect this.

enricovianello (approved):

PR #957, opened 2025-04-17
Avoid persisting too long request parameters

garaimanoj:
Upgraded dependencies:
- mitreid to 1.3.7.cnaf-20250404
- voms-api-java to 3.3.5


enricovianello:
I merged the mitreID pull request, we can consider also this PR as status:done

enricovianello:
I also managed to upgrade VOMS API Java version here https://github.com/indigo-iam/iam/pull/966 because it was not only a version upgrade but I had to remove/change other dependencies related (bouncy castle in primis). But I think If we merge this PR before we won't have particular problems.

PR #959, opened 2025-04-17
Access tokens are no longer saved to the database

### LINKED ISSUE #537: Option needed for IAM not to store access tokens in the DB

enricovianello:
Related to these changes on MitreID https://github.com/indigo-iam/OpenID-Connect-Java-Spring-Server/pull/25

First implementation details:

- compute signature check each time an access token is received
- introspection endpoint returns only info stored on token
- persist on database the revoked access tokens
- cache the validation result of an access token


enricovianello:
To be reopened/refactored after the merge of https://github.com/indigo-iam/iam/pull/1041

PR #967, opened 2025-04-29
Fix bug related to group managers

rmiccoli:
In particular, group manager was no longer able to see information on the members of the groups they manage, because the endpoint `/iam/multi-factor-settings/{accountId}` - showing the mfa status of the account - is accessible only by admins.


enricovianello (approved):
Waiting to change the permissions on MFA settings, this PR simply avoid not showing anything to the group manager just for an error code from the API.

PR #969, opened 2025-04-30
SCIM API bulk operations

### LINKED ISSUE #935: SCIM API bulk operations
Currently, IAM doesn't support [bulk operations](https://datatracker.ietf.org/doc/html/rfc7644#section-3.7). The only way to create/patch users in bulk is sending a request per user (this could create so many requests on a script that synchronizes user information in a big VO). It would be more efficient if IAM implements the SCIM bulk operations.

Context:
At CERN's CMS instance we synchronize the group memberships via a cronjob. To create new users, we have to send the requests one by one for each new user. And even worse, we need to send a patch request for each existing user to update their certificate. If we had a bulk operations endpoint for the Users, it would be sending 2 requests instead of ~7k request.

I believe this would need to be added as a new endpoint. e.g. something like `/scim/Users/Bulk` sounds like a good candidate to me. And it would need to handle 2 request types: e.g. `POST /scim/Users/Bulk` to create users in bulk and `PATCH /scim/Users/Bulk` to patch users in bulk.

And it would be great if the documentation can be updated accordingly once this is implemented.

mmutadich:


mmutadich:
The maximum number of operations that the request will take has been set to an arbitrary value so far (500) and the default failOnErrors value is -1 to allow any number of errors. Please let me know if you have any suggestions for more appropriate values!

iam-persistence/pom.xml
```diff
@@ -44,6 +44,11 @@
       <artifactId>spring-boot-starter-validation</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-validation</artifactId>
+    </dependency>
+
```
enricovianello: It seems this dependency has been added as a duplicate

enricovianello (changes requested):
Just a duplication to be removed

enricovianello (approved):

PR #972, opened 2025-05-02
Notify VO Admins when a certificate is added by a user

### LINKED ISSUE #961: Notify VO Admins when a certificate is added by a user
It was requested by ALICE VO Admins to get notified when a user adds a certificate to their account.

- This feature needs to be configurable by each instance and should default to not notifying the admins to keep backwards compatibility (e.g. can be set in the application.yaml or env variables)
- Every time a user adds a certificate to their account, an email needs to be sent to the VO Admins ( depending on the IAM_NOTIFICATION_ADMIN_NOTIFICATION_POLICY=notify-admins)

S203932:
Added a new environment variable in the 'application.yml' for the login service called `certificateUpdate` and it has the following default value `false`, i.e. `${IAM_NOTIFICATION_CERTIFICATE:false}`.

When enabled admins will be notified, when there is a new certificate added to the user and when there is a certificate removed from a user.


S203932:
In need of review and input on the newly created method `getNewValue` in the file `iam-login-service/src/main/java/it/infn/mw/iam/api/scim/updater/DefaultAccountUpdater.java`. 
See the method `createLinkedCertificateMessage` in `iam-login-service/src/main/java/it/infn/mw/iam/notification/TransientNotificationFactory.java` for implementation of the `getNewValue` method.

enricovianello:
Conflicts to be fixed

enricovianello (approved):

iam-login-service/src/main/java/it/infn/mw/iam/api/account_linking/DefaultAccountLinkingService.java
```diff
@@ -187,25 +197,55 @@ public void linkX509Certificate(Principal authenticatedUser,
               userAccount.getUsername(), x509Credential.getSubject()),
           x509Credential));
     }
+
+    notificationProperties.getAdminNotificationPolicy();
```
enricovianello: probably the idea here was to save this value in a variable

iam-login-service/src/main/java/it/infn/mw/iam/api/account_linking/DefaultAccountLinkingService.java
```diff
@@ -187,25 +197,55 @@ public void linkX509Certificate(Principal authenticatedUser,
...
+    if (Boolean.TRUE.equals(notificationProperties.getCertificateUpdate())
+        && (notificationProperties.getAdminNotificationPolicy()
```
enricovianello: I'm not understanding the reason why the notification policy is evaluated here. The policy could be admin|address|admin_and_address but this doesn't mean that the email has not to be sent. Seeing how the factory works, probably this policy should be evaluated inside the factory logic, as it's done for example here https://github.com/indigo-iam/iam/blob/develop/iam-login-service/src/main/java/it/infn/mw/iam/notification/TransientNotificationFactory.java#L169-L171

iam-login-service/src/main/java/it/infn/mw/iam/api/account_linking/DefaultAccountLinkingService.java
```diff
@@ -187,25 +197,55 @@ public void linkX509Certificate(Principal authenticatedUser,
...
+
+    IamX509Certificate certificate = null;
+
+    Set<IamX509Certificate> certificates = userAccount.getX509Certificates();
+    for (IamX509Certificate cert : certificates) {
+      if (cert.getSubjectDn().equals(certificateSubject)) {
+        certificate = cert;
+        break;
+      }
+    }
```
enricovianello, line 221: we can still use Java 8+ streams and Optional:

```
  Optional<IamX509Certificate> certificate = userAccount.getX509Certificates().stream()
    .filter(cert -> cert.getSubjectDn().equals(certificateSubject) && cert.getIssuerDn().equals(certificateIssuer))
    .findFirst();
``` 
and we must check both subject and issuer.
Then:

```
if (certificate.isPresent()) {
  removed = userAccount.getX509Certificates().remove(certificate.get());
}

File: iam-login-service/src/main/java/it/infn/mw/iam/api/account_linking/DefaultAccountLinkingService.java [hunk 4]
```diff
@@ -187,25 +197,55 @@ public void linkX509Certificate(Principal authenticatedUser,
...
           "User '%s' unlinked certificate with subject '%s' and issuer '%s' from his/her membership",
           userAccount.getUsername(), certificateSubject, certificateIssuer), certificateSubject,
           certificateIssuer));
+
+      if (Boolean.TRUE.equals(notificationProperties.getCertificateUpdate())
+          && (notificationProperties.getAdminNotificationPolicy()
+            .equals(AdminNotificationPolicy.NOTIFY_ADMINS)
+              || notificationProperties.getAdminNotificationPolicy()
+                .equals(AdminNotificationPolicy.NOTIFY_ADDRESS_AND_ADMINS))) {
+        notificationFactory.createUnlinkedCertificateMessage(userAccount, certificate);
```
enricovianello: same comment as the one above

iam-login-service/src/main/java/it/infn/mw/iam/notification/NotificationFactory.java
```diff
@@ -65,4 +68,15 @@ IamEmailNotification createClientStatusChangedMessageFor(ClientDetailsEntity cli
   IamEmailNotification createSetAsServiceAccountMessage(IamAccount account);
 
   IamEmailNotification createRevokeServiceAccountMessage(IamAccount account);
+
+  IamEmailNotification createLinkedCertificateMessage(IamAccount account, AccountUpdater u);
+
+  IamEmailNotification createLinkedCertificateMessage(IamAccount account,
```
enricovianello, line 70: Can we implement only a unique createLinkedCertificateMessage with the following args:
* IamAccount account (the account owner of the certificate)
* IamX509AuthenticationCredential x509Credential (the certificate)
?

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/updater/AccountUpdater.java
```diff
@@ -20,4 +20,6 @@
 public interface AccountUpdater extends Updater {
 
   public IamAccount getAccount();
+
+  public <T> T getNewValue(Class<T> clazz);
```
enricovianello: We're adding this just to have issuer + subject DN of the linked/unlinked certificate. I would try something different. It seems to me a very complex implementation, just for logging/mailing reasons.

iam-login-service/src/main/java/it/infn/mw/iam/notification/TransientNotificationFactory.java
```diff
@@ -508,4 +510,142 @@ protected IamEmailNotification createMessage(String templateName, Map<String, Ob
       return null;
     }
   }
+
+
+
+  @Override
```
enricovianello: Cosmetic fix: I'd place these methods above the "createMessage" one

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/updater/AccountUpdater.java
enricovianello: For this reason I'm suggesting that we need to find a way to implement a unique createLinkedCertificateMessage method described below :)

enricovianello (changes requested):
I'm not really convinced by the "newValue" strategy and we could probably apply a simpler approach, if possible. I'd try to change x509Certificate [Adders](https://github.com/indigo-iam/iam/blob/master/iam-login-service/src/main/java/it/infn/mw/iam/api/scim/updater/builders/Adders.java#L209-L214) and [Removers](https://github.com/indigo-iam/iam/blob/master/iam-login-service/src/main/java/it/infn/mw/iam/api/scim/updater/builders/Removers.java#L81-L87) implementation with a dedicated object ([as done for username](https://github.com/indigo-iam/iam/blob/master/iam-login-service/src/main/java/it/infn/mw/iam/api/scim/updater/builders/Replacers.java#L139-L143)) in order to override the [afterUpdate method](https://github.com/indigo-iam/iam/blob/master/iam-login-service/src/main/java/it/infn/mw/iam/api/scim/updater/DefaultUpdater.java#L80-L82) and put there the logic of the mail notification (at that point you have access to account and also the old and new value). I'd give a try to this approach :)

iam-login-service/src/main/java/it/infn/mw/iam/api/account_linking/DefaultAccountLinkingService.java
S203932: Sorry, it was leftover from debugging. It had no function, it has been removed :+1:

iam-login-service/src/main/java/it/infn/mw/iam/api/account_linking/DefaultAccountLinkingService.java
S203932: I re-read the issue and realize that the notification should only be based upon the value of `updateCertificate`. 
I've removed the evaluation of the policy, but kept the evaluation of `certificateUpdate` as I understand the notification should only be created given that this value is `true`. 

Regarding where the evaluation happens, do you want me to call the method `createUnlinkedCertificateMessage`/ `createLinkedCertificateMessage` in either case and then do the evaluation of the `certificateUpdate` value from the method `createUnlinkedCertificateMessage`/ `createLinkedCertificateMessage`?

iam-login-service/src/main/java/it/infn/mw/iam/api/account_linking/DefaultAccountLinkingService.java
S203932: Nice, thanks, I've changed it accordingly :smile:

iam-login-service/src/main/java/it/infn/mw/iam/api/account_linking/DefaultAccountLinkingService.java
S203932: I've changed it similarly to above :+1:

iam-login-service/src/main/java/it/infn/mw/iam/notification/NotificationFactory.java
S203932: It has been done. 
Now `createLinkedCertificateMessage` and `createUnlinkedCertificateMessage` has the parameters `IamAccount` and ` IamX509AuthenticationCredential`. 
They're defined as follows:

    IamEmailNotification createLinkedCertificateMessage(IamAccount account,
            IamX509AuthenticationCredential x509Credential);

    IamEmailNotification createUnlinkedCertificateMessage(IamAccount account,
            IamX509AuthenticationCredential x509Credential);

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/updater/AccountUpdater.java
S203932: The method `getNewValue` has been removed and is no longer necessary :+1:

iam-login-service/src/main/java/it/infn/mw/iam/notification/TransientNotificationFactory.java
S203932: I've moved it around :+1:

CHANGELOG.md
```diff
@@ -72,6 +72,7 @@ It will be enhanced and expanded in future releases, based also on users feedbac
 
 * Fix scope policy filter when refresh token has been already issued by @enricovianello in https://github.com/indigo-iam/iam/pull/982
 
+
```
enricovianello: I know I could be seen as a bad person but .. this line change is useless :)

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
```diff
@@ -546,4 +570,13 @@ public void update(final String id, final List<ScimPatchOperation<ScimUser>> ope
     operations.forEach(op -> executePatchOperation(account, op));
   }
 
+  private IamX509AuthenticationCredential getNewX509AuthCredValue(ScimIndigoUser indigoUser) {
```
enricovianello: You could also create a "converter" method inside the ScimX509Certificate object, for example something like:

```
  public IamX509AuthenticationCredential asIamX509AuthenticationCredential() {
    return new IamX509AuthenticationCredential.Builder().issuer(getIssuerDn())
      .subject(getSubjectDn())
      .build();
  }
  ```
 So this conversion could be easily used somewhere else :)

File: iam-login-service/src/main/java/it/infn/mw/iam/api/scim/updater/AccountUpdater.java [hunk 10]
```diff
@@ -20,4 +20,5 @@
 public interface AccountUpdater extends Updater {
 
   public IamAccount getAccount();
+
```
enricovianello: I'm always the same bad person :)

iam-login-service/src/main/java/it/infn/mw/iam/notification/NotificationFactory.java
```diff
@@ -28,41 +29,47 @@
 
 public interface NotificationFactory {
 
-  IamEmailNotification createConfirmationMessage(IamRegistrationRequest request);
+    IamEmailNotification createConfirmationMessage(IamRegistrationRequest request);
```
enricovianello: this indentation seems not following the 2 spaces.. strange

enricovianello (changes requested):
Really small optimizations required. Thanks for the changes after the previous review :) I think this solution is good and after these final optimizations it will be merged

CHANGELOG.md
S203932: No worries, it has been removed :smile:

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
S203932: It has been added :+1:

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/updater/AccountUpdater.java
S203932: No worries, I've removed it as well :smile:

iam-login-service/src/main/java/it/infn/mw/iam/notification/NotificationFactory.java
S203932: The file has now been re-formatted, can you confirm?

iam-login-service/src/main/java/it/infn/mw/iam/api/account_linking/DefaultAccountLinkingService.java
enricovianello: I think current implementation is fine (`certificateUpdate` outside the factory methods) 👍

enricovianello (approved):

PR #973, opened 2025-05-02
Move acr logic out from mitre and OIDC auth provider refactoring

enricovianello:


iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/IamOAuth2RequestFactory.java
```diff
@@ -264,10 +268,32 @@ private Map<String, String> updatedTokenRequestParameters(
 
     });
 
+    validateAndUpdateAcrRequest(tokenRequestParameters);
```
rmiccoli, line 271: I don't think this is the right place to request the acr value. We need to request/propagate it during the authentication request (before login), not at the token endpoint.

enricovianello: You're right :)

PR #974, opened 2025-05-07
Add endpoint to retrieve clients owned by user

### LINKED ISSUE #962: Add endpoint to retrieve the list of clients owned by a user
From the API documentation, there is the `/iam/account/me/clients` endpoint which provides the list of clients owned by the user itself.
It is missing its counterpart `/iam/account/{id}/clients` to let an admin to retrieve the list of clients owned by a different user.

garaimanoj:
image


iam-login-service/src/main/java/it/infn/mw/iam/api/account/client/AccountClientController.java
```diff
@@ -52,4 +53,14 @@ public ListResponseDTO<RegisteredClientDTO> getOwnedClients(@Validated Paginated
     return clientSearchService.findOwnedClients(form);
   }
 
+  @JsonView(ClientViews.Limited.class)
+  @PreAuthorize("hasRole('ADMIN')")
```
enricovianello, line 57: I think we should authorize as done in others endpoint in this way:

`#iam.hasScope('iam:admin.read') or #iam.hasDashboardRole('ROLE_ADMIN') or #iam.isUser(#id)`

enricovianello: This probably means also changes on tests.

enricovianello (changes requested):
I think that endpoint can be accessed also by the owner of the clients

iam-login-service/src/main/java/it/infn/mw/iam/api/account/client/AccountClientController.java
garaimanoj: Done

rmiccoli (approved):

enricovianello (approved):

PR #975, opened 2025-05-08

garaimanoj:


rmiccoli (changes requested):
The authZ on the GET endpoint is ok, but I am not sure about the PATCH, as it is now possible for a user to change their username, status, etc. It differs from what is is allowed with the PATCH on the `scim/Me` endpoint.

garaimanoj:
Replaced by https://github.com/indigo-iam/iam/pull/980

PR #978, opened 2025-05-09
Revoke all refresh tokens for a client

### LINKED ISSUE #577: FR: Revoke all refresh tokens for a client
Feature request :)

In the case of a security incident we would need to revoke all refresh tokens for a client. 

Some users may wish to revoke access tokens as well (though revocation of access tokens is obviously up for debate...)

mmutadich:


iam-login-service/src/main/java/it/infn/mw/iam/api/client/management/ClientManagementAPIController.java
```diff
@@ -157,6 +171,28 @@ public void disableClient(@PathVariable String clientId) {
     account.ifPresent(a -> managementService.updateClientStatus(clientId, false, a.getUuid()));
   }
 
+  @PatchMapping("/{clientId}/revoke-refresh-tokens")
+  @PreAuthorize("#iam.hasScope('iam:admin.write') or #iam.hasDashboardRole('ROLE_ADMIN')")
+  public void revokeRefreshTokens(@PathVariable String clientId) {
+    disableClient(clientId);
```
enricovianello: I can understand disabling the client seems reasonable but in my opinion we could keep this operation separated. Reading the last comment of Hannah here https://github.com/indigo-iam/iam/issues/577#issuecomment-2245410776 I can understand that the flow is "I found a compromised client, I rotated the secret and I want to revoke all the issued tokens for a security reason". So probably you don't want to disable it. And I think this is not expected if I call an endpoint called "revoke-refresh-tokens". So, I'd remove this line.

iam-login-service/src/main/java/it/infn/mw/iam/api/client/management/ClientManagementAPIController.java
```diff
@@ -157,6 +171,28 @@ public void disableClient(@PathVariable String clientId) {
...
+    ClientDetailsEntity client = clientService.findClientByClientId(clientId)
+      .orElseThrow(ClientSuppliers.clientNotFound(clientId));
+    tokenService.getRefreshTokensForClient(client).forEach(tokenService::revokeRefreshToken);
+    rotateClientSecret(clientId);
```
enricovianello: Same as above. There's another endpoint dedicated to this. I'd prefer to have a "revoke-refresh-token" and another endpoint called "resetClient" that does "disable-revoke-rotate-enable" flow. But this endpoint needs to return the new secret. We're moving in a direction that the secret is returned once it has been created and then hashed into the db.

enricovianello (commented):
I have some concerns on the implementation. The revoke endpoint does too many things in my opinion and I'd keep it limited to the expected outcome "all the tokens are revoked". But I understand the need so I also would like to understand if it makes sense to have a dedicated "{clientId}/reset" endpoint that does the flow "disable-rotate-revoke-enable" flow.

iam-login-service/src/main/webapp/resources/iam/apps/dashboard-app/components/clients/client/status/client.status.component.js
```diff
@@ -47,39 +47,86 @@
...
+    self.disableClient = function () {
       return ClientsService.disableClient(self.client.client_id)
-          .then(self.handleSuccess)
-          .catch(self.handleError);
+        .then(self.handleSuccess)
+        .catch(self.handleError);
     };
 
+    self.removeTokens = function () {
+      return ClientsService.removeTokens(self.client.client_id)
```
enricovianello: I cannot find this method: there are 2 defined methods: revokeRefreshTokens and revokeAccessTokens instead. Probably this could be a wrapper for both of them or you need another button + function

enricovianello (changes requested):
I think there's something (easy) to fix on the Angular dashboard but I didn't run it. But I'm quite sure about :)

iam-login-service/src/main/webapp/resources/iam/apps/dashboard-app/components/clients/client/status/client.status.component.js
mmutadich: I have removed this because it is implemented in the wrong file - now the button logic should be in client.component.js using the following methods: revokeTokens and resetClient!

iam-login-service/src/test/java/it/infn/mw/iam/test/api/TestSupport.java
```diff
@@ -19,75 +19,111 @@
...
 public class TestSupport {
 
-  public static final ResultMatcher OK = status().isOk();
-  public static final ResultMatcher NO_CONTENT = status().isNoContent();
-  public static final ResultMatcher BAD_REQUEST = status().isBadRequest();
-  public static final ResultMatcher UNAUTHORIZED = status().isUnauthorized();
-  public static final ResultMatcher FORBIDDEN = status().isForbidden();
-  public static final ResultMatcher NOT_FOUND = status().isNotFound();
-  public static final ResultMatcher CREATED = status().isCreated();
+    public static final ResultMatcher OK = status().isOk();
```
enricovianello: strange indentation. Should be 2 spaces but it doesn't seem

iam-login-service/src/test/java/it/infn/mw/iam/test/api/client/ClientManagementAPIIntegrationTests.java
```diff
@@ -46,283 +50,421 @@
...
-      .andExpect(jsonPath("$.active").value(true));
-
-    mvc.perform(patch(ClientManagementAPIController.ENDPOINT + "/{clientId}/disable", "client")
-    ).andExpect(OK);
-
-    mvc.perform(get(ClientManagementAPIController.ENDPOINT + "/client"))
-      .andExpect(OK)
-      .andExpect(jsonPath("$.active").value(false));
-  }
+        @Autowired
```
enricovianello: very strange indentation also here

enricovianello (changes requested):
Just a cosmetic indentation fix. I need to test it from the point of view of the UI but the PR is approved in the end. Thanks :)

enricovianello (approved):

PR #980, opened 2025-05-22

### LINKED ISSUE #964: GET/PATCH of a user from /scim/Me and /scim/Users/{id} should be equivalent if the users are themselves
As title

GET/PATCH of a user from `/scim/Me` and `/scim/Users/{id}` should be equivalent if the user doing the operations if the users are themselves

garaimanoj:


garaimanoj:
@rmiccoli :The authZ on the GET endpoint is ok, but I am not sure about the PATCH, as it is now possible for a user to change their username, status, etc. It differs from what is is allowed with the PATCH on the scim/Me endpoint.

rmiccoli (approved):
This PR looks good to me :)

garaimanoj:
Due to merge conflict, replaced by [PR](https://github.com/indigo-iam/iam/pull/1010)

PR #981, opened 2025-05-22
Token lifetime configurable in token request

### LINKED ISSUE #869: Token lifetime configurable in token request
There seems to be an increasing demand to provide a configurable token lifetime based on a parameter in the initial token request. For example, perhaps a scope could be used to modify the lifetime. This would allow tools such as Rucio to tune the lifetime based on the risk profile of the process. There would need to be a maximum set for the client to mitigate risk. This issue is to start a small discussion on whether it is a useful feature, whether it is somehow supported in the standard already, and how it could be done.

mmutadich:


iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/common/BaseAccessTokenBuilder.java
```diff
@@ -126,10 +127,17 @@ protected boolean hasAudienceRequest(OAuth2Authentication authentication) {
     return !isNullOrEmpty(audience);
   }
 
+  protected boolean hasCustomValidityRequest(OAuth2Authentication authentication) {
+    final String audience = authentication.getOAuth2Request().getRequestParameters().get(EXPIRES_IN_KEY);
```
federicaagostini, line 131: Rename variable (something like `expiration`).

iam-login-service/src/test/java/it/infn/mw/iam/test/oauth/TokenLifetimeConfigurableTests.java
```diff
@@ -0,0 +1,139 @@
...
+    JWT token = JWTParser.parse(accessToken);
+
+    JWTClaimsSet claims = token.getJWTClaimsSet();
+
+    assertNotNull(claims.getIssueTime());
+    assertNotNull(claims.getExpirationTime());
+    assertThat((int) (claims.getExpirationTime().getTime() - claims.getIssueTime().getTime()) / 1000, equalTo(maxValidity));
+  }
+
+}
```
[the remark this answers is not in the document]
mmutadich: I have implemented a test that generates two refresh tokens, one where the access token lifetime is configured using the "expires_in" parameter in the request and the other without this parameter. The test then executes the refresh token flow using both tokens and verifies if the resulting access tokens have the expected lifetimes. I wanted to confirm with you if my understanding of the expected lifetimes is correct. If the 'expires_in' parameter was provided in the refresh request, then access tokens issued using this refresh token should have that configured lifetime. Otherwise access tokens should have the lifetime defined by the client. Please let me know if this is correct @federicaagostini !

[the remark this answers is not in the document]
enricovianello: In my opinion this is the expected way it should work 👍

enricovianello (approved):

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/common/BaseAccessTokenBuilder.java
```diff
@@ -126,10 +128,18 @@ protected boolean hasAudienceRequest(OAuth2Authentication authentication) {
     return !isNullOrEmpty(audience);
   }
 
+  protected boolean hasCustomValidityRequest(OAuth2Authentication authentication) {
+    final String audience =
```
federicaagostini: Rename variable (e.g. `expiration`)

iam-login-service/src/test/java/it/infn/mw/iam/test/oauth/TokenLifetimeConfigurableTests.java
```diff
@@ -0,0 +1,262 @@
...
+import it.infn.mw.iam.api.common.client.RegisteredClientDTO;
+import it.infn.mw.iam.test.util.annotation.IamMockMvcIntegrationTest;
+import it.infn.mw.iam.core.oauth.profile.common.BaseAccessTokenBuilder;
+
+@SuppressWarnings("deprecation")
+@RunWith(SpringRunner.class)
+@IamMockMvcIntegrationTest
+public class TokenLifetimeConfigurableTests {
+
+        public static final String TEST_USERNAME = "test";
```
federicaagostini, line 59: Please use formatter

iam-login-service/src/test/java/it/infn/mw/iam/test/oauth/TokenLifetimeConfigurableTests.java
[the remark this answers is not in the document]
federicaagostini: Hi, sorry for the very late replay. What I expect is that the refresh token is not bounded to any expiration, i.e.
* during an access token request using also the `offline_access` scope and `expires_in` parameter:
  * the access token should have a lifetime equal to the issue time + expires_in
  * the refresh token should have a the lifetime configured at client level (usually 30 days)
* during a refresh token request using the previously issued refresh token:
  * if the `expires_in` parameter is requested, the access token should have the custom lifetime
  * if the `expires_in` parameter is not requested, the access token should have the lifetime configured at client level (usually 3600 s).

At least this is what I would expect, but let's involve other people in the discussion, since there is no spec (as far as I know) about the `expires_in` feature.

iam-login-service/src/test/java/it/infn/mw/iam/test/oauth/TokenLifetimeConfigurableTests.java
```diff
@@ -0,0 +1,262 @@
...
+      .andReturn()
+      .getResponse()
+      .getContentAsString();
+    // @formatter:on
+
+                DefaultOAuth2AccessToken ordinaryToken = mapper.readValue(ordinaryTokenResponse,
+                                DefaultOAuth2AccessToken.class);
+
+                String ordinaryRefresh = ordinaryToken.getRefreshToken().toString();
+
```
federicaagostini, line 314: I would check here that the AT has a 3600s lifetime.

iam-login-service/src/test/java/it/infn/mw/iam/test/oauth/TokenLifetimeConfigurableTests.java
```diff
@@ -0,0 +1,262 @@
...
+                DefaultOAuth2AccessToken tokenResponse = mapper
+                        .readValue(configuredLifetimeTokenResponse, DefaultOAuth2AccessToken.class);
+
```
federicaagostini, line 322: I would check that the AT has the lifetime equal to the `expires_in` parameter

iam-login-service/src/test/java/it/infn/mw/iam/test/oauth/TokenLifetimeConfigurableTests.java
```diff
@@ -0,0 +1,262 @@
...
+                String refreshwithConfiguredAccessToken =
+                                tokenResponse.getRefreshToken().toString();
```
federicaagostini: Check that the RT lifetime is 30 days

iam-login-service/src/test/java/it/infn/mw/iam/test/oauth/TokenLifetimeConfigurableTests.java
federicaagostini, line 322: Check that the RT lifetime is 30 days

iam-login-service/src/test/java/it/infn/mw/iam/test/oauth/TokenLifetimeConfigurableTests.java
```diff
@@ -0,0 +1,262 @@
...
+
+    // @formatter:off
+    String ordinaryReponse = mvc.perform(post("/token")
+        .with(httpBasic(clientId, clientSecret))
+        .param("grant_type", "refresh_token")
+        .param("refresh_token", ordinaryRefresh))
```
federicaagostini: Please add `expires_in` parameter and check the AT lifetime

iam-login-service/src/test/java/it/infn/mw/iam/test/oauth/TokenLifetimeConfigurableTests.java
```diff
@@ -0,0 +1,262 @@
...
+                String configuredAccessToken =
+                                mapper.readTree(configuredReponse).get("access_token").asText();
+                JWTClaimsSet configuredClaims =
+                                JWTParser.parse(configuredAccessToken).getJWTClaimsSet();
+
+                assertNotNull(configuredClaims.getIssueTime());
+                assertNotNull(configuredClaims.getExpirationTime());
+                assertThat((int) (configuredClaims.getExpirationTime().getTime()
+                                - configuredClaims.getIssueTime().getTime()) / 1000,
+                                equalTo(Integer.parseInt(CUSTOM_LIFETIME)));
```
federicaagostini: In my personal view this should fail (I would expect it is equal to 3600)

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/granters/IamRefreshTokenGranter.java
```diff
@@ -65,7 +70,16 @@ protected OAuth2AccessToken getAccessToken(ClientDetails client, TokenRequest to
...
     }
 
-    return getTokenServices().refreshAccessToken(refreshTokenValue, tokenRequest);
+    OAuth2AccessToken newToken =
+        getTokenServices().refreshAccessToken(refreshTokenValue, tokenRequest);
+    
+    if (tokenRequest.getRequestParameters().containsKey("expires_in")){
+      OAuth2Authentication authentication = new OAuth2Authentication(
+        getRequestFactory().createOAuth2Request(client, tokenRequest), null);
+      newToken = tokenEnhancer.enhance(newToken, authentication);
```
enricovianello: Re-creating the authentication object always hides some consequences but luckily the framework will help us :)
In fact you can check directly inside the BaseAccessTokenBuilder logic you've already updated if there is a `EXPIRES_IN_KEY` within the requested parameter (done) or if the authentication object `isRefresh` and there is that parameter within `getRefreshTokenRequest()` object (to-do).

We are sure the refresh mechanism is used from these lines:
https://github.com/indigo-iam/OpenID-Connect-Java-Spring-Server/blob/v1.3.7/openid-connect-server/src/main/java/org/mitre/oauth2/service/impl/DefaultOAuth2ProviderTokenService.java#L396-L397

This means that we're adding to the original request the "refreshed" one that can be retrieved with the method `getRefreshTokenRequest()` and checked with `isRefresh()`.

> This is the refresh implementation from Spring framework:
>
> ```
>	public OAuth2Request refresh(TokenRequest tokenRequest) {
>		OAuth2Request request = new OAuth2Request(getRequestParameters(), getClientId(), authorities, approved,
>				getScope(), resourceIds, redirectUri, responseTypes, extensions);
>		request.refresh = tokenRequest;
>		return request;
>	}
>```

I'd give a try to this approach, it seems well integrated with the framework and we're not overwriting the old requests.

enricovianello (changes requested):
I've suggested a slight different approach. Very similar to your but I think well integrated with the framework

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/common/BaseAccessTokenBuilder.java
```diff
@@ -152,6 +164,14 @@ protected JWTClaimsSet.Builder baseJWTSetup(OAuth2AccessTokenEntity token,
       audience = authentication.getOAuth2Request().getRequestParameters().get(AUD_KEY);
     }
 
+    if (hasCustomValidityRequest(authentication)) {
+      expiry = authentication.getOAuth2Request().getRequestParameters().get(EXPIRES_IN_KEY);
+    }
+
+    if(hasCustomValidityRefreshFlow(authentication)){
```
enricovianello: If I understood well how this framework works, probably you can check this condition (and understand if it's a refresh-flow) or else check the previous one. Doing this, it's unnecessary to filter the `expires_in` parameter inside IamTokenService.

```
if (hasCustomValidityRefreshFlow(authentication)) {
  expiry = authentication.getOAuth2Request().getRefreshTokenRequest().getRequestParameters().get(EXPIRES_IN_KEY);
} else if (hasCustomValidityRequest(authentication)) {
  expiry = authentication.getOAuth2Request().getRequestParameters().get(EXPIRES_IN_KEY);
}

**Review by enricovianello — CHANGES_REQUESTED** [review]
Probably we can optimize something more and avoid filtering the requested parameters inside IamTokenService

File: iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/common/BaseAccessTokenBuilder.java [hunk 12]
```diff
@@ -152,6 +174,12 @@ protected JWTClaimsSet.Builder baseJWTSetup(OAuth2AccessTokenEntity token,
       audience = authentication.getOAuth2Request().getRequestParameters().get(AUD_KEY);
     }
 
+    if (isRefreshFlow(authentication)){
+      expiry = getRefreshFlowExpiration(authentication, token);
+    } else if (hasCustomValidityRequest(authentication)) {
+      expiry = authentication.getOAuth2Request().getRequestParameters().get(EXPIRES_IN_KEY);
+    }
+
```
enricovianello: If I understood well the code of 
`org.springframework.security.oauth2.provider.OAuth2Request.refresh()`
the "refresh" token request is stored inside refresh field and we can access it through `authentication.getOAuth2Request().getRefreshTokenRequest()`
while the original access token request (+ offline_access) can be accessed through `authentication.getOAuth2Request()`.

Probably we can define a unique method to compute the final expiration.

```
Date expTime = computeExpTime(authentication, token);
```

```
private Date computeExpTime(OAuth2Authentication authentication, OAuth2AccessTokenEntity token) {

  OAuth2Request originalRequest = authentication.getOAuth2Request();
  if (originalRequest.isRefresh()) {
    OAuth2Request refreshRequest = originalRequest.getRefreshTokenRequest();
    // check if current request has the expires_in
    if (refreshRequest.getRequestParameters().containsKey(EXPIRES_IN_KEY)) {
      return min(refreshRequest.getRequestParameters().get(EXPIRES_IN_KEY), token.getClient().getAccessTokenValiditySeconds()); // or something like min
    }
    if (originalRequest.getRequestParameters().containsKey(EXPIRES_IN_KEY)) {
      return min(originalRequest.getRequestParameters().get(EXPIRES_IN_KEY), token.getClient().getAccessTokenValiditySeconds()); // or something like min
    }
    return token.getClient().getAccessTokenValiditySeconds();
}
```

I think this could work. If your test is green it should be easy to check.

enricovianello (changes requested):
Probably we can simplify even more

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/common/BaseAccessTokenBuilder.java
```diff
@@ -163,6 +183,20 @@ protected JWTClaimsSet.Builder baseJWTSetup(OAuth2AccessTokenEntity token,
...
+      expiry = computeExpTime(authentication, token);
+    } catch (NumberFormatException e) {
+      throw new InvalidRequestException(INVALID_PARAMETER);
+    }
+
+    if (!isNull(expiry) && expiry >= 0){
+      expTime = Date.from(issueTime.plus(expiry, ChronoUnit.SECONDS));
+      token.setExpiration(expTime);
+    }
+
```
enricovianello: having the default value in `token.getExpiration()` we could have here only few lines like these:

```
Date expTime = computeExpTime(authentication, token);
token.setExpiration(expTime);
builder.expirationTime(expTime);
```

In practice, we could move both null and positive number check and the fallback to the `token.getExpiration()` value (in case of invalid values or exceptions) inside the `computeExpTime` method.

But

another approach that now I realize could be even better is to move this logic from here to IamTokenEnhancer.enhance() method. Then, here we could assume the token.getExpiration() has been already computed so we can directly assign that value to the JWT claim.

Inside IamTokenEnhancer.enhance() instead, we can compute the token expiration time (so we don't need to update the entity later) and we can do it because we have access to all the things we need: from OAuth2Authentication and OAuth2AccessTokenEntity objects to ClientDetailsEntity settings.

enricovianello: One approach could be only moving the logic inside enhance() method and apply it for all the clients as it is now. Default value: not enabled. Inside next release we can add it as configurable per client.

mmutadich: I moved the logic to IamTokenEnhancer.enhance() but we can discuss making it client configurable tomorrow!

enricovianello (approved):

PR #984, opened 2025-06-11
Configurable linking of certificate upon registration request

### LINKED ISSUE #968: allow user registration to link X509 certificate already
Currently, the user registration essentially has 2 parts:
* the initial registration, to be approved by a VO admin;
* the linking of a certificate to the IAM account.
In some VOs, there still is work to be done by a VO admin when the certificate DN has been added.
While #961 already helps, it would be better if the user could link the certificate already during the registration.
That is how things worked in VOMS-Admin.

S203932:
One can now configure if the user must, may or can not connect their X509 certificate to their registration request. 
This can be configured in the `application.yml` via the environment variable `require-certificate`.
The 3 options are `required`, `optional` and `off`. 

The registration page changes layout accordingly. 
Font has been changed for the errors displayed when the certificate is required. 

Pop-up window with warning appears when there is a month or less before the expiration date of the certificate.


S203932:
In need of review. 

The sonar cube issue is not related to changes I made, but is only registered as new code, as I formatted all the files I changed (and this part wasn't consistent with the desired format).

The same goes for 2 of the 4 security Hotspots detected by Sonar Cloud. 
Please advice how to attend these. 

The feature of this pull request is front-end heavy and has quite a few changes to the registration page depending on the value of the environment variable `require-certificate`.

iam-login-service/src/test/java/it/infn/mw/iam/test/registration/RegistrationRequestCertificateTests.java
```diff
@@ -0,0 +1,255 @@
...
+                    .orElseThrow(() -> new AccountNotFoundException(
+                            "Can not remove suspended account as none is found"));
+
+        for (IamRegistrationRequest iamRegistrationRequest : requests) {
+            iamRequestRepo.delete(iamRegistrationRequest);
+        }
+
+        IamAccount account =
+                iamAccountRepo.findByUsername(USERNAME).orElseThrow(AccountNotFoundException::new);
+
```
rmiccoli, line 165: Here I would make a further check, i.e. before deleting the account I would verify that the account contains that certificate, e.g.
```
 assertTrue(account.getX509Certificates().stream()
        .anyMatch(c -> c.getCertificate().contentEquals(TEST_0_CERT)));
```

and for completeness I would also test the other cases, especially `iam.registration.require-certificate=OPTIONAL` +  `request.setRegisterCertificate("true/false");`.

iam-login-service/src/main/java/it/infn/mw/iam/registration/DefaultRegistrationRequestService.java
```diff
@@ -126,6 +135,13 @@ public class DefaultRegistrationRequestService
   @Autowired
   private IamProperties iamProperties;
 
+  private final X509CertificateConverter x509Converter;
+
+  @Autowired
```
rmiccoli: I would simply add
```
@Autowired
private X509CertificateConverter x509Converter;
```

iam-login-service/src/main/java/it/infn/mw/iam/authn/x509/IamX509PreauthenticationProcessingFilter.java
```diff
@@ -109,10 +118,25 @@ protected Object getPreAuthenticatedPrincipal(HttpServletRequest request) {
...
       return null;
     }
 
+
+    Date expirationDate = credential.get().getCertificateChain()[0].getNotAfter();
+    Calendar calendar = Calendar.getInstance();
+    calendar.add(Calendar.MONTH, 1);
+    Date minTimeBeforeExpiration = calendar.getTime();
+    if (expirationDate.before(minTimeBeforeExpiration)) {
+      request.setAttribute(X509_ALMOST_EXPIRED, Boolean.TRUE);
```
rmiccoli, line 138: So does this apply even if the certificate has already expired?

iam-login-service/src/main/java/it/infn/mw/iam/registration/DefaultRegistrationRequestService.java
S203932: Hi @rmiccoli,
I previously had the implementation you suggest, but this was flagged by Sonar Cloud as an issue, so I went with the current implementation (see commit 41be87dfd75acced16e76de8343c9b814612e1c5).
Do you want me to change it back and keep the Sonar Cloud issue?

iam-login-service/src/test/java/it/infn/mw/iam/test/registration/RegistrationRequestCertificateTests.java
S203932: Hi @rmiccoli, I've added additional tests that should cover what you've requested. 
Can you verify that the tests cover what you've requested and if not, could you suggest what further tests that might be desired?

iam-login-service/src/main/java/it/infn/mw/iam/authn/x509/IamX509PreauthenticationProcessingFilter.java
S203932: Hi @rmiccoli, 
Given that a certificate has expired its header should have the `X-SSL-Client-Verify` value as "FAILED: certificate has expired" (see https://nginx.org/en/CHANGES-1.12 under section 
Changes with nginx 1.11.7). From this the method `parseVerifyHeader` in `DefaultX509AuthenticationCredentialExtractor` would return the `X509CertificateVerificationResult` as failed (line 97). This would be returned to the `extractX509Credential` within the same file that builds and returns the `IamX509AuthenticationCredential` object. The  method `extractCredential` in `IamX509PreauthenticationProcessingFilter` calls the method `extractX509Credential` to get a hold of the `IamX509AuthenticationCredential` object and checks if the x509 credentials are still valid (line 100). If it is not valid (i.e. if the X509 is expired) then the method `extractCredential` returns an `Optional.empty()` (line 103).
This `Optional.empty()` is given to the method `getPreAuthenticatedPrincipal` within the `IamX509PreauthenticationProcessingFilter` and since it is not present in case the certificate is expired, the method would return `null` at line 127.

Thus, it should not show the pop-up window to the user about the "1 month or less to certificate expiration", in case the certificate has already expired. 

Does this answer the question or do you have other concerns?

iam-login-service/src/test/java/it/infn/mw/iam/test/registration/RegistrationRequestCertificateTests.java
rmiccoli: Yes, these tests are enough :)

iam-login-service/src/main/java/it/infn/mw/iam/registration/DefaultRegistrationRequestService.java
rmiccoli: No, don't worry, you can leave it like that. I just didn't like the difference with the other _autowired_

iam-login-service/src/main/java/it/infn/mw/iam/authn/x509/IamX509PreauthenticationProcessingFilter.java
rmiccoli: All clear, thanks!

rmiccoli:
> ## [Quality Gate Passed](https://sonarcloud.io/dashboard?id=indigo-iam_iam&pullRequest=984) **Quality Gate passed**
> 
> Issues  [1 New issue](https://sonarcloud.io/project/issues?id=indigo-iam_iam&pullRequest=984&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [0 Accepted issues](https://sonarcloud.io/project/issues?id=indigo-iam_iam&pullRequest=984&issueStatuses=ACCEPTED)
> 
Try to solve this sonar issue by replacing the type `boolean` (for the `doNotSendEmail` parameter) with the `Boolean` wrapper.

S203932:
> > ## [Quality Gate Passed](https://sonarcloud.io/dashboard?id=indigo-iam_iam&pullRequest=984) **Quality Gate passed**
> > Issues  [1 New issue](https://sonarcloud.io/project/issues?id=indigo-iam_iam&pullRequest=984&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [0 Accepted issues](https://sonarcloud.io/project/issues?id=indigo-iam_iam&pullRequest=984&issueStatuses=ACCEPTED)
> 
> Try to solve this sonar issue by replacing the type `boolean` (for the `doNotSendEmail` parameter) with the `Boolean` wrapper.

@rmiccoli, the Sonar issue has now been fixed

rmiccoli (approved):

rmiccoli:
I approved this PR, but it needs a rebase from develop branch

S203932:
> I approved this PR, but it needs a rebase from develop branch

It should now have been rebased from `develop` branch. @rmiccoli can you verify?

rmiccoli:
> > I approved this PR, but it needs a rebase from develop branch
> 
> It should now have been rebased from `develop` branch. @rmiccoli can you verify?

I still see conflicts, I'll try to help you with the rebase :)

iam-login-service/src/main/java/it/infn/mw/iam/config/IamProperties.java
```diff
@@ -40,7 +40,7 @@ public enum EditableFields {
   }
 
   public enum RegistrationField {
-    EMAIL, NAME, SURNAME, USERNAME, AFFILIATION, NOTES
+    EMAIL, NAME, SURNAME, USERNAME, AFFILIATION, NOTES, CERTIFICATE
```
enricovianello, line 43: I cannot find any usage of this CERTIFICATE field. Is it used by the ui? Anyway adding a field here it means there's the opportunity to map it to a value that comes from a remote provider, it adds also a behavior (default is mandatory, you can change it within application-registration.yaml).
If it's not really necessary, we can remove this.

iam-login-service/src/main/java/it/infn/mw/iam/config/IamProperties.java
```diff
@@ -55,6 +55,10 @@ public enum LocalAuthenticationLoginPageMode {
     VISIBLE, HIDDEN, HIDDEN_WITH_LINK
   }
 
+  public enum RequireCertificateOption {
+    OFF, OPTIONAL, REQUIRED
```
enricovianello: isn't OPTIONAL enough to cover the OFF case? OFF doesn't mean the user is not able to link his certificate later so I'd evaluate (to simplify the implementation) if it's really necessary or not.

iam-login-service/src/main/java/it/infn/mw/iam/config/security/IamWebSecurityConfig.java
```diff
@@ -165,15 +169,15 @@ protected void configure(final HttpSecurity http) throws Exception {
 
       // @formatter:off
       http.requestMatchers()
-        .antMatchers("/", "/login**", "/logout", "/authorize", "/manage/**", "/dashboard**",
+        .antMatchers("/", "/login**", "/logout", AUTHORIZE, "/manage/**", "/dashboard**",
```
enricovianello: Was this a Sonar suggestion? XD I think we could keep using several "/authorize" strings. The best solution should be change all of them (e.g. creating a "Endpoints" class with all the constants) but it could be a separated task.

iam-login-service/src/main/java/it/infn/mw/iam/config/IamProperties.java
enricovianello: Probably OFF makes sense to not bother users to be prompted to provide a certificate when approaching the registration page. isn' it?

iam-login-service/src/main/java/it/infn/mw/iam/config/IamProperties.java
enricovianello: Or you can rely on its "behavior" as a registration field (MANDATORY, OPTIONAL, HIDDEN) to implement the logic without adding another parameter. I know this could be a big change (maybe not) to this code but it gives sense to the fact that:
1. it's true that CERTIFICATE is a registration field
2. gives sense to its inherited behavior sub-field
3. doesn't create strange configuration (mandatory CERTIFICATE vs OFF certificate option)
4. less parameters

iam-login-service/src/main/java/it/infn/mw/iam/registration/DefaultRegistrationRequestService.java
```diff
@@ -309,7 +342,39 @@ private RegistrationRequestDto handleReject(IamRegistrationRequest request,
...
   }
 
+  private void linkRequestCertificateToAccount(IamAccount account, HttpServletRequest request) {
 
+    HttpSession session = request.getSession(false);
+
+    IamX509AuthenticationCredential cred = Optional
+      .ofNullable(
+          (IamX509AuthenticationCredential) session.getAttribute(X509_CREDENTIAL_SESSION_KEY))
+      .orElseThrow(() -> new IllegalArgumentException("No X.509 credential found in session "));
```
enricovianello, line 365: this case is not managed from the caller so we probably end up in an unfinished registration (AupSignature is not created for example).
We could probably log the error and go on. User could retry later to link his certificate. Otherwise a well managed failure (with no account creation) has to be implemented. A second registration attempt wouldn't be possibile without the manual delete of the current reg-request/user. Isn't it?

iam-login-service/src/main/java/it/infn/mw/iam/registration/DefaultRegistrationRequestService.java
```diff
@@ -309,7 +342,39 @@ private RegistrationRequestDto handleReject(IamRegistrationRequest request,
...
+    final Date now = Date.from(clock.instant());
+    cert.setAccount(account);
+    cert.setLabel("cert-0");
+    cert.setPrimary(true);
+    cert.setCreationTime(now);
+    cert.setLastUpdateTime(now);
+
+    iamAccountRepo.findByCertificateSubject(cert.getSubjectDn()).ifPresent(c -> {
+      iamAccountRepo.delete(account);
+      throw new CredentialAlreadyBoundException(
```
enricovianello: same consideration about errors as above. What happens in this case? I'd prefer a "rollback solution". Otherwise these things could be evaluated in a "validation" phase that comes before creating the account, the linked certificate, the aup signature, etc.

enricovianello: the initial validation check could simplify the management of the errors

enricovianello (changes requested):
It's ok but I think something could be optimized. We can have a dedicated call in case we want to discuss something.

iam-login-service/src/main/java/it/infn/mw/iam/config/security/IamWebSecurityConfig.java
S203932: Hahah yep, trying to minimize Sonar issues, but I've changed it back :+1:

iam-login-service/src/main/java/it/infn/mw/iam/config/IamProperties.java
S203932: The `require-certificate` environment variable has been removed. 
Now it is just relying on the behavior from registration field :+1:

iam-login-service/src/main/java/it/infn/mw/iam/registration/DefaultRegistrationRequestService.java
S203932: Sanity checks have been implemented instead of this :+1:

iam-login-service/src/main/java/it/infn/mw/iam/registration/DefaultRegistrationRequestService.java
S203932: Sanity checks have been implemented instead of this :+1: 
In case of error, no account will be created and the registration fails

enricovianello (changes requested):
Conflicts to be fixed

enricovianello (approved):

PR #994, opened 2025-06-19
Option to limit custom scopes to admins only

### LINKED ISSUE #895: Feature Request: option to limit new scope definition to admins
Could we please get an option, VO/instance wide, to restrict defining new scopes to admins (or group of users)? Right now users can define new scopes at will for any of their clients.
Thanks,
- Stephan

mmutadich:
This PR adds configurable support for restricting **custom OAuth/OIDC client scopes to administrators**. When enabled, non-admin users can no longer add or retain custom scopes during client registration or updates, while administrators remain unrestricted.

High-level changes:

- **Backend authorization**
  - Added filtering of custom scopes during client registration and client updates.
  - Applied the restriction only to non-admin users when the feature is enabled.
  - Preserved existing handling of reserved and restricted system scopes.
  - Ensured custom scopes cannot be reintroduced through client updates.

- **Configuration**
  - Added the `admin-only-custom-scopes` client-registration property.
  - Exposed it through the existing configuration model and environment-variable configuration:
    - `IAM_CLIENT_ADMIN_ONLY_CUSTOM_SCOPES`
  - The setting can be enabled or disabled without code changes.

- **Frontend behavior**
  - Added the configuration flag to the login-page configuration exposed to JavaScript.
  - Hid the custom-scope controls from non-admin users when admin-only mode is active.
  - Kept the controls visible to administrators.


iam-login-service/src/main/webapp/resources/iam/apps/dashboard-app/components/clients/client/scopelist/scopelist.component.html
```diff
@@ -45,7 +45,8 @@
                 </div>
             </div>
         </div>
-        <div class="form-group">
+        <div class="form-group"
+            ng-show="($root.loggedUser.isAdmin && $root.isAdminOnlyCustomScope) || (!$root.isAdminOnlyCustomScope)">
```
rmiccoli: I think it should be moved before `<label>Custom scopes</label>` to also hide **Custom scopes** from normal users.

rmiccoli (approved):

enricovianello (approved):

PR #997, opened 2025-06-23
Configurable text in the registration button

### LINKED ISSUE #995: Configurable "Apply for an account" button text
It would be good to have the "Apply for an account" button's text configurable in the .yaml or env (or both), just like we have for the OIDC button. Maybe it can be configurable with a new config value under iam.registration in the application.yaml or somewhere similar. The default value should be the current text to have backwards compatibility

Rationale:
In some of our CERN deployments, we allow registration with only a CERN SSO account or eduGAIN (via CERN SSO), and we would like to mention this in the button by changing it to something like "Register with eduGAIN or CERN SSO" because otherwise it creates confusion in the users.

S203932:
The text is configurable through the through the environment variable `IAM_REGISTRATION_BUTTON_TEXT` in the `application.yml` file. 

Default value is "Apply for an account"

### DIFF
--- a/iam-login-service/src/main/webapp/WEB-INF/tags/iam/page.tag
+++ b/iam-login-service/src/main/webapp/WEB-INF/tags/iam/page.tag
@@ -101,6 +101,10 @@
 	function getExternalAuthenticationEnabled() {
 		return ${loginPageConfiguration.externalAuthenticationEnabled};
 	}
+
+  function getRegistrationButtonText() {
+    return '${loginPageConfiguration.registrationButtonText}';
+  }
 	
 	function getOrganisationName() {
 		return '${iamOrganisationName}';
--- a/iam-login-service/src/main/webapp/WEB-INF/tags/iam/page.tag
+++ b/iam-login-service/src/main/webapp/WEB-INF/tags/iam/page.tag
@@ -102,9 +102,9 @@
 		return ${loginPageConfiguration.externalAuthenticationEnabled};
 	}
 
-  function getRegistrationButtonText() {
-    return '${loginPageConfiguration.registrationButtonText}';
-  }
+  	function getRegistrationButtonText() {
+    		return '${loginPageConfiguration.registrationButtonText}';
+  	}
 	
 	function getOrganisationName() {
 		return '${iamOrganisationName}';
--- a/iam-login-service/src/main/webapp/WEB-INF/views/iam/login.jsp
+++ b/iam-login-service/src/main/webapp/WEB-INF/views/iam/login.jsp
@@ -191,7 +191,7 @@
                 <div class="registration-preamble text-muted">
                    Not a member?
                 </div>
-                <a class="btn btn-success btn-block" href="/start-registration">Apply for an account</a>
+            <a class="btn btn-success btn-block" href="/start-registration">${loginPageConfiguration.registrationButtonText}</a>
                 <c:if test="${
                     loginPageConfiguration.samlEnabled
                     && iamSamlProperties.wayfLoginButton.visible
--- a/iam-login-service/src/main/webapp/WEB-INF/views/iam/login.jsp
+++ b/iam-login-service/src/main/webapp/WEB-INF/views/iam/login.jsp
@@ -191,14 +191,14 @@
                 <div class="registration-preamble text-muted">
                    Not a member?
                 </div>
-            <a class="btn btn-success btn-block" href="/start-registration">${loginPageConfiguration.registrationButtonText}</a>
-                <c:if test="${
-                    loginPageConfiguration.samlEnabled
-                    && iamSamlProperties.wayfLoginButton.visible
-                    && iamSamlProperties.wayfLoginButton.image.url.contains('edugain-logo.gif')
-                }">
-                  <!-- WAYF login button -->
-                  <a class="btn btn-success btn-block" href="/saml/login">Register an account with eduGAIN</a>
+                    <a class="btn btn-success btn-block" href="/start-registration">${loginPageConfiguration.registrationButtonText}</a>
+                        <c:if test="${
+                            loginPageConfiguration.samlEnabled
+                            && iamSamlProperties.wayfLoginButton.visible
+                            && iamSamlProperties.wayfLoginButton.image.url.contains('edugain-logo.gif')
+                        }">
+                    <!-- WAYF login button -->
+                    <a class="btn btn-success btn-block" href="/saml/login">Register an account with eduGAIN</a>
                 </c:if>
             </div>
         </c:if>
--- a/iam-login-service/src/test/java/it/infn/mw/iam/test/login/RegistrationButtonDisabledTests.java
+++ b/iam-login-service/src/test/java/it/infn/mw/iam/test/login/RegistrationButtonDisabledTests.java
@@ -15,12 +15,12 @@
  */
 package it.infn.mw.iam.test.login;
 
-import static org.hamcrest.CoreMatchers.containsString;
-import static org.hamcrest.CoreMatchers.not;
+import static org.junit.Assert.assertEquals;
 
 import org.junit.BeforeClass;
 import org.junit.Test;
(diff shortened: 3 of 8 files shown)


iam-login-service/src/test/java/it/infn/mw/iam/test/login/RegistrationButtonDisabledTests.java
```diff
@@ -29,34 +29,59 @@
...
+    while (responseBody.indexOf(iamProperties.getRegistration().getRegistrationButtonText(),
+        index) != -1) {
+      amountOccurences++;
+      index =
+          responseBody.indexOf(iamProperties.getRegistration().getRegistrationButtonText(), index)
+              + 1;
+
+    }
+
+    assertEquals(1, amountOccurences);
```
rmiccoli, line 78: I would also add `assertTrue(responseBody.contains("Apply for an account"));`

iam-login-service/src/test/java/it/infn/mw/iam/test/login/RegistrationButtonTextTests.java
```diff
@@ -0,0 +1,91 @@
...
+import it.infn.mw.iam.config.IamProperties;
+import it.infn.mw.iam.test.TestUtils;
+import it.infn.mw.iam.test.util.annotation.IamRandomPortIntegrationTest;
+
+
+
+@RunWith(SpringRunner.class)
+@IamRandomPortIntegrationTest
+@SpringBootTest(classes = {IamLoginService.class}, webEnvironment = WebEnvironment.RANDOM_PORT)
+@TestPropertySource(properties = {"iam.registration.show-registration-button-in-login-page=true"})
```
rmiccoli: Here, instead, I would change the value of `iam.registration.registration-button-text` property and verify that `assertTrue(responseBody.contains("Another value"));`

iam-login-service/src/test/java/it/infn/mw/iam/test/login/RegistrationButtonDisabledTests.java
S203932: I'm assuming this is because I should not assume a dynamically updated value of `registrationButtonText`, but rather assume the default value when doing testing?

If that is the case, then I can make a local variable `private String REGISTRATIONBUTTONTEXT = "Apply for an account";` 

And when counting amount of instances I can do the following:
```

    while (responseBody.indexOf(REGISTRATIONBUTTONTEXT, index) != -1) {
      amountOccurences++;
      index = responseBody.indexOf(REGISTRATIONBUTTONTEXT, index) + 1;

    }

    assertEquals(1, amountOccurences);
    assertTrue(responseBody.contains(REGISTRATIONBUTTONTEXT));
```
Though the `assertTrue` is a bit superfluous as the value of `amountOccurences` is to show how many instances of the value `REGISTRATIONBUTTONTEXT` there is, right?
Is the `assertTrue`  necessary?

iam-login-service/src/test/java/it/infn/mw/iam/test/login/RegistrationButtonTextTests.java
S203932: So if I'm understanding correctly I should assume the default values in the `application.yml` and not write them explicitly (i.e. remove `@TestPropertySource(properties = {"iam.registration.show-registration-button-in-login-page=true"}) ` as the default value is `true`).

If that is the case I'll change the line with:
`@TestPropertySource(properties = {"iam.registration.registration-button-text=Another value"})`

And then I can add a variable `private String REGISTRATIONBUTTONTEXT = "Another value";`

and change the test to the following:

```
        while (responseBody.indexOf(REGISTRATIONBUTTONTEXT, index) != -1) {
            amountOccurences++;
            index = responseBody.indexOf(REGISTRATIONBUTTONTEXT, index) + 1;

        }

        assertEquals(2, amountOccurences);
        assertTrue(responseBody.contains(REGISTRATIONBUTTONTEXT));
```
Though again, I'm not sure the `assertTrue` is necessary, given that amountOccurences checks for the amount of appearances of the `REGISTRATIONBUTTONTEXT`.

rmiccoli: > So if I'm understanding correctly I should assume the default values in the `application.yml` and not write them explicitly (i.e. remove `@TestPropertySource(properties = {"iam.registration.show-registration-button-in-login-page=true"}) ` as the default value is `true`).

yes, that's right!

> If that is the case I'll change the line with: `@TestPropertySource(properties = {"iam.registration.registration-button-text=Another value"})`
> 
> And then I can add a variable `private String REGISTRATIONBUTTONTEXT = "Another value";`
> 
> and change the test to the following:
> 
> ```
>         while (responseBody.indexOf(REGISTRATIONBUTTONTEXT, index) != -1) {
>             amountOccurences++;
>             index = responseBody.indexOf(REGISTRATIONBUTTONTEXT, index) + 1;
> 
>         }
> 
>         assertEquals(2, amountOccurences);
>         assertTrue(responseBody.contains(REGISTRATIONBUTTONTEXT));
> ```
> 
> Though again, I'm not sure the `assertTrue` is necessary, given that amountOccurences checks for the amount of appearances of the `REGISTRATIONBUTTONTEXT`.


I understand this is the purpose of the test, but I would also like to test the value of the property when a value other than the default is given. Perhaps another test can be written?

iam-login-service/src/test/java/it/infn/mw/iam/test/login/RegistrationButtonDisabledTests.java
S203932: I've now added a new test, does this fulfill what you requested @rmiccoli ?

iam-login-service/src/test/java/it/infn/mw/iam/test/login/RegistrationButtonTextTests.java
S203932: I've now added a new test, does this fulfill what you requested @rmiccoli ?

rmiccoli (approved):

iam-login-service/src/main/webapp/WEB-INF/tags/iam/page.tag
```diff
@@ -101,6 +101,10 @@
 	function getExternalAuthenticationEnabled() {
 		return ${loginPageConfiguration.externalAuthenticationEnabled};
 	}
+
+  function getRegistrationButtonText() {
+    return '${loginPageConfiguration.registrationButtonText}';
+  }
```
enricovianello: Indentation to fix

iam-login-service/src/main/webapp/WEB-INF/views/iam/login.jsp
```diff
@@ -191,7 +191,7 @@
                 <div class="registration-preamble text-muted">
                    Not a member?
                 </div>
-                <a class="btn btn-success btn-block" href="/start-registration">Apply for an account</a>
+            <a class="btn btn-success btn-block" href="/start-registration">${loginPageConfiguration.registrationButtonText}</a>
```
enricovianello: also here it seems indentation is wrong

enricovianello (changes requested):
Just some indentation fixes

iam-login-service/src/main/webapp/WEB-INF/tags/iam/page.tag
S203932: Should have been fixed now, though can only verify this when editing the file.
@enricovianello can you confirm that the indentation is fixed?

iam-login-service/src/main/webapp/WEB-INF/views/iam/login.jsp
S203932: I've changed the formatting of select code. 
@enricovianello  can you confirm this is the desired formatting?

enricovianello (approved):

PR #998, opened 2025-06-24
Configurable support button

### LINKED ISSUE #996: Configurable support button
It would be good to have a configurable support button with a configurable link that can be added to the login page. It could be just under the `Privacy Policy` link with the same format and configurable in the same way (e.g. just like IAM_PRIVACY_POLICY_URL env variable) To keep backwards compatibility, the default should be no button/link.

Rationale: At CERN, we want to add a link on the login page directing users to open a support ticket, to assist users who don't know how to contact us.

mmutadich:


rmiccoli (approved):
Looks good to me!

enricovianello (approved):
But some merge conflicts needs to be fixed

PR #1005, opened 2025-07-08
Update affiliation with HR DB

### LINKED ISSUE #755: institute not displayed to VO Admin role
I see in the code an `institute` var returned [here](https://github.com/indigo-iam/iam/blob/19633a38c418f4f467f8779650032bc45e321843/iam-login-service/src/main/java/it/infn/mw/iam/api/registration/cern/dto/ParticipationDTO.java#L37), which I assume says what it means, but as a VO admin (ATLAS), I do not see this value displayed in user records, even for my own membership. Is there a reason this field is not shown to VO admins (or possibly more or all roles), or has it been hidden intentionally? I think admins may need to see this value when troubleshooting certain membership issues -- it is certainly a common factor in current VOMS issues.

mmutadich:


enricovianello (dismissed):

iam-login-service/src/main/java/it/infn/mw/iam/core/lifecycle/cern/CernHrLifecycleHandler.java
```diff
@@ -126,7 +126,10 @@ public void handleAccount(String cernPersonId, String experiment, IamAccount a)
         expireAccount(a);
       }
       return;
+    } else {
+      a.getUserInfo().setAffiliation(ep.get().getInstitute().getName());
     }
+
```
enricovianello, line 131: you can simply put line 130 after the if because if it's empty it will return in the end.

enricovianello (changes requested):
A very very small cosmetic fix. I know I can be a bad person :) It's approved!

iam-login-service/src/main/java/it/infn/mw/iam/core/lifecycle/cern/CernHrLifecycleHandler.java
mmutadich: Good catch! :)

enricovianello (approved):

PR #1010, opened 2025-07-16

### LINKED ISSUE #964: GET/PATCH of a user from /scim/Me and /scim/Users/{id} should be equivalent if the users are themselves

garaimanoj:


garaimanoj:
Copy of [PR-980](https://github.com/indigo-iam/iam/pull/980)

rmiccoli (approved):

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
```diff
@@ -542,8 +556,75 @@ private void handleSpecificUpdateType(IamAccount account, AccountUpdater u) {
...
 
-    operations.forEach(op -> executePatchOperation(account, op));
+    if (!updatesToPublish.isEmpty()) {
+      account.touch();
+      accountRepository.save(account);
+      updatesToPublish.forEach(u -> u.publishUpdateEvent(this, eventPublisher));
+    }
+  }
+
+  public Set<UpdaterType> getEnabledUpdaters() {
```
enricovianello: This method could be static I think

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
```diff
@@ -542,8 +556,75 @@ private void handleSpecificUpdateType(IamAccount account, AccountUpdater u) {
...
+
+    iamProperties.getUserProfile().getEditableFields().forEach(e -> {
+      if (EditableFields.NAME.equals(e)) {
+        enabledUpdaters.add(ACCOUNT_REPLACE_GIVEN_NAME);
+      } else if (EditableFields.SURNAME.equals(e)) {
+        enabledUpdaters.add(ACCOUNT_REPLACE_FAMILY_NAME);
+      } else if (EditableFields.PICTURE.equals(e)) {
+        enabledUpdaters.add(ACCOUNT_REPLACE_PICTURE);
+        enabledUpdaters.add(ACCOUNT_REMOVE_PICTURE);
+      } else if (EditableFields.EMAIL.equals(e)) {
```
enricovianello: This has not been introduced by you but I'm thinking about giving the opportunity to the user to change the email address: we should send an email to verify it! Probably we should avoid letting users to change their email (by default) and we should work on introducing a replace mechanism that verifies the new email. Not so easy.

enricovianello (changes requested):
Just a cosmetic fix. And a reflection about letting users to change their email but this topic cannot be managed here :)

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
enricovianello: by default email is included in the editable fields: I'd remove it honestly https://github.com/indigo-iam/iam/blob/develop/iam-login-service/src/main/resources/application.yml#L166

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
garaimanoj: If we make it static then accessing `iamProperties.getUserProfile()` is not possible without passing `iamProperties` as a parameter. In that case we need instance of `IamProperties` in `ScimMeController` and `ScimUserProvisioning`.

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
```diff
@@ -564,7 +577,75 @@ else if (ACCOUNT_REMOVE_X509_CERTIFICATE.equals(u.getType())) {
...
     IamAccount account = accountRepository.findByUuid(id).orElseThrow(() -> noUserMappedToId(id));
+    Optional<IamAccount> currentUserAccount = getCurrentUserAccount();  
+    if (!currentUserAccount.isPresent() || isAdmin(currentUserAccount.get())) {
+      operations.forEach(op -> executePatchOperation(account, op));
+    } else {
+      operations.forEach(op -> executePatchOperationByUser(account, op));
+    }
+  }
+
+  private boolean isAdmin(IamAccount userAccount) {
```
rmiccoli: You can use the `isAdmin()` method already present in `AccountUtils` class (https://github.com/indigo-iam/iam/blob/develop/iam-login-service/src/main/java/it/infn/mw/iam/api/account/AccountUtils.java#L60)

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
```diff
@@ -564,7 +577,75 @@ else if (ACCOUNT_REMOVE_X509_CERTIFICATE.equals(u.getType())) {
   public void update(final String id, final List<ScimPatchOperation<ScimUser>> operations) {
 
     IamAccount account = accountRepository.findByUuid(id).orElseThrow(() -> noUserMappedToId(id));
+    Optional<IamAccount> currentUserAccount = getCurrentUserAccount();
```
rmiccoli: You can use https://github.com/indigo-iam/iam/blob/develop/iam-login-service/src/main/java/it/infn/mw/iam/api/account/AccountUtils.java#L87

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
garaimanoj: Done

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
garaimanoj: Done

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
enricovianello: True

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
```diff
@@ -564,7 +575,54 @@ else if (ACCOUNT_REMOVE_X509_CERTIFICATE.equals(u.getType())) {
   public void update(final String id, final List<ScimPatchOperation<ScimUser>> operations) {
 
     IamAccount account = accountRepository.findByUuid(id).orElseThrow(() -> noUserMappedToId(id));
+    Optional<IamAccount> currentUserAccount = accountUtils.getAuthenticatedUserAccount();  
+    if (!currentUserAccount.isPresent() || accountUtils.isAdmin(currentUserAccount.get())) {
```
jacogasp: I would suggest to add an assert to be 110% sure that at this point the role `ROLE_ADMIN` is present, or make it a little bit clearer.
Without digging through the call stack, this condition looks like an unauthenticated user has the same power as the admin.

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/controller/ScimMeController.java
rmiccoli, line 1: We should add the following to the `executePatchOperation` method
```
if (!accountUtils.isAdmin(account) && !enabledUpdaters.contains(u.getType())) {
        throw new ScimPatchOperationNotSupported(u.getType().getDescription() + " not supported");
}
```
because we have noticed that if, for example, we remove `name` from the configuration
```
user-profile:
    editable-fields:
    - email
    - name
    - picture
    - surname
```
neither the user nor the admin can change their name, but admin should be able to do so.

garaimanoj: Done

iam-login-service/src/main/java/it/infn/mw/iam/api/scim/provisioning/ScimUserProvisioning.java
garaimanoj: Code refactored.

rmiccoli (approved):

iam-login-service/src/test/java/it/infn/mw/iam/test/scim/user/ScimUserProvisioningPatchTests.java
```diff
@@ -433,4 +438,27 @@ public void testAddPicture() throws Exception {
...
+  public void testUserCanChangeEmail() throws Exception {
+    ScimUser updates = ScimUser.builder()
+        .addEmail(ScimEmail.builder()
+            .type(ScimEmailType.home)
+            .email("TestUser@example.com")
+            .primary(false).build())
+        .build();
+
+    scimUtils.patchUser(lennon.getId(), replace, updates, HttpStatus.NO_CONTENT)
+        .andExpect(status().isNoContent());
```
enricovianello, line 508: I would add a check if the user's email has really changed after the operation

enricovianello (changes requested):
It seems ready to be merged. Some conflicts need to be fixed and I'd add a final check on the test. If a test with an admin user that uses the scim/Me endpoint to update some of his fields is not present, I would add it too.

iam-login-service/src/test/java/it/infn/mw/iam/test/scim/user/ScimUserProvisioningPatchTests.java
garaimanoj: Done

enricovianello (approved):

PR #1013, opened 2025-07-21
Ensure removal of configured MFA totp before account deletion

garaimanoj:


rmiccoli (approved):

iam-login-service/src/main/java/it/infn/mw/iam/core/user/DefaultIamAccountService.java
```diff
@@ -237,11 +241,22 @@ protected void deleteTokensForAccount(IamAccount account) {
     }
   }
 
+  private void deleteTotpMfa(IamAccount account) {
+    Optional<IamTotpMfa> totpMfaOptional = iamTotpMfaRepository.findByAccount(account);
+    if (totpMfaOptional.isPresent()) {
+      IamTotpMfa totpMfa = totpMfaOptional.get();
+      iamTotpMfaRepository.delete(totpMfa);
+
+      saveAccount(account);
```
enricovianello: Why is this necessary?

enricovianello (commented):
Only one clarification needed

iam-login-service/src/main/java/it/infn/mw/iam/core/user/DefaultIamAccountService.java
garaimanoj: `Account` doesn't hold any reference of `IamTotpMfa`  so I was not able to apply `cascade = CascadeType.REMOVE`

garaimanoj: `saveAccount(account);` removed

enricovianello (approved):

PR #1017, opened 2025-07-29
Configurable 'nbf' duration for tokens

### LINKED ISSUE #704: Handling Access Token Refusals Due to Time Discrepancies
We've encountered an issue where some access tokens are occasionally rejected by the Computing Elements (CEs). The primary error observed on the CE side is shown below:

```
[2024-01-29 10:43:53] [Arc.JWSE] [DEBUG] [20077/6] JWSE::Input: JWS content: {
        ...
        "nbf":  1706521434,  // Mon Jan 29 2024 10:43:54 GMT+0100
        ...
}
[2024-01-29 10:43:53] [Arc.JWSE] [DEBUG] [20077/6] JWSE::Input: JWS: token too young
...
[2024-01-29 10:43:53] [Arc.MCC.HTTP] [WARNING] [20077/6] HTTP Error: 403 User can't be assigned configuration
```

The error, "token too young", indicates a problem with the `nbf` (Not Before) claim. It appears the CE's current time (`2024-01-29 10:43:53`) is slightly behind the `nbf` time of the token (`1706521434 = Mon Jan 29 2024 10:43:54 GMT+0100`).

Further evidence of time discrepancy is observed in the CE's logs:

```
[2024-01-29 10:22:01] [Arc.validator] [WARNING] Timecheck: Your time differs slightly (1.604302 seconds) from the public time server 'europe.pool.ntp.org'.
```

Although the root cause lies with the time synchronization on the CE side, we propose a pragmatic solution to enhance the robustness of our system: configuring a small leeway in the `nbf` claim of the access tokens. This approach aligns with the guidelines in the JWT RFC (https://www.rfc-editor.org/rfc/rfc7519#section-4.1.5), which states:

> Implementers MAY provide for some small leeway, usually no more than a few minutes, to account for clock skew.

By introducing this leeway, we aim to mitigate token rejection issues stemming from minor time differences without compromising security.

We're keen to hear thoughts and feedback on this approach, particularly concerning best practices and potential implications.

mmutadich:
Resolves #704


iam-login-service/src/main/java/it/infn/mw/iam/config/IamProperties.java
```diff
@@ -467,6 +460,8 @@ public static class AccessToken {
     boolean includeAuthnInfo = false;
     boolean includeScope = false;
     boolean includeNbf = false;
+    int maxNbf = 300;
```
enricovianello: This property in the end could be configured but it exists just to be an hard coded maximum value that validates the value of the other property (`nbfOffsetSeconds`). Solutions could be:
1. remove this property and accept anything the administrator/deployer sets
2. hardcode this value in order to validate the value set by the administrator/deployer
In my opinion the option 1 is enough, simpler and 100% dynamic in the hands of the deployers :)

enricovianello (changes requested):
I would remove the maxNbf (or hardcode its value)

iam-login-service/src/main/java/it/infn/mw/iam/config/IamProperties.java
mmutadich: I have removed the property!

iam-login-service/src/test/java/it/infn/mw/iam/test/oauth/NbfTokenTests.java
```diff
@@ -51,9 +61,64 @@ public void testScopeIncludedInAccessTokenClientCred() throws Exception {
...
+    assertThat(token.getJWTClaimsSet().getNotBeforeTime(),
+        equalTo(Date.from(token.getJWTClaimsSet()
+          .getIssueTime()
+          .toInstant()
+          .minus(Duration.ofSeconds(properties.getAccessToken().getNbfOffsetSeconds())))));
+
+  }
+
+  @Test
+  public void testScopeNotIncludedInAccessTokenClientCred() throws Exception {
```
enricovianello: Probably a cut & paste problem :) here the name probably was something like testNbfNotIncluded... and not scope

enricovianello (changes requested):
Just a very cosmetic fix. It's approved anyway

enricovianello (approved):

PR #1019, opened 2025-07-30
Resolve local login flow with hidden usecase

### LINKED ISSUE #859: When local login is hidden, sll=y no longer works
The documentation (https://indigo-iam.github.io/v/v1.10.0/docs/reference/configuration/local_authn/) specifies that the sll=y link should work even if the local login page is hidden as done below.

```
iam:
  local-authn:
    # possible values for enabled-for: all, vo-admins, none
    enabled-for: vo-admins 
    # possible values: hidden, hidden-with-link, visible
    login-page-visibility: hidden 
```

This is no longer true and the sll=y url becomes unavailable. 
We would like to be able to completely hide this option on the login page but be able to access it if absolutely needed (e.g. for disaster recovery).

Sae126V:
Resolves #859

### DIFF
--- a/iam-login-service/src/main/webapp/WEB-INF/views/iam/login.jsp
+++ b/iam-login-service/src/main/webapp/WEB-INF/views/iam/login.jsp
@@ -68,7 +68,7 @@
             test="${
                 (
                   loginPageConfiguration.localAuthenticationVisible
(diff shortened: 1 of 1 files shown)


Sae126V:
Raised another PR from https://github.com/indigo-iam/iam/pull/1053

PR #1025, opened 2025-08-05
VO-wide config to disable token-exchange upscoping

### LINKED ISSUE #1020: Prevent "upscoping" in the token-exchange workflow
As discussed at the IAM Hackathon (July 2025), following up with a ticket on this topic. The request is to have a way to prevent clients from obtaining additional scopes (upscoping) via the `token-exchange` workflow.

During the `token-exchange`, the acting client should not be able to obtain scopes outside of those present in the subject token. This is particularly relevant for FTS, where we are interested in obtaining a refresh token, but purposefully don't change token scopes, as we want to conserve the ones from the initial subject token.

**Steps to reproducing**

1. Obtain the initial (subject) token
```bash
$ SUBJECT_TOKEN=$(...)
$ token-decode "${SUBJECT_TOKEN}" | jq .scope
"storage.read:/"
```

2. Perform the token exchange
```bash
$ curl -s -u "${CLIENT_ID}:${CLIENT_SECRET}" \
       -H "Content-Type: application/x-www-form-urlencoded" \
       --data "grant_type=urn:ietf:params:oauth:grant-type:token-exchange" \
       --data "subject_token_type=urn:ietf:params:oauth:token-type:access_token" \
       --data "subject_token=${SUBJECT_TOKEN}" \
       --data "scope=storage.read:/ storage.modify:/ offline_access" \
       --data "audience=https://wlcg.cern.ch/jwt/v1/any" \
    https://dteam-auth.cern.ch/token
```

3. Exchanged token contains additional scopes
```bash
$ EXCHANGED_TOKEN=$(..)
$ token-decode "${EXCHANGED_TOKEN}" | jq .scope
"offline_access storage.modify:/ storage.read:/""
```

**Request**

The token-exchange should not allow obtaining additional scopes not present in the initial, subject token (with the obvious exception of the `offline_access`).

I understand this change will not be desirable for all communities, which is why this setting is a good candidate to be a per-client configuration option. At least for WLCG + FTS, we'd prefer upscoping is not allowed, limiting a vulnerable pathway for the FTS usecase.

mmutadich:


iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/exchange/DefaultTokenExchangePdp.java
```diff
@@ -99,14 +109,29 @@ private TokenExchangePdpResult verifyScopes(TokenExchangePolicy p, TokenRequest
...
-    // The requested scopes must be allowed by the origin client (destination is impersonating the
-    // origin client)
-    Set<ScopeMatcher> originClientMatchers = scopeMatcherRegistry.findMatchersForClient(origin);
+    Set<ScopeMatcher> scopeMatchers = scopeMatcherRegistry.findMatchersForClient(origin);
+    String invalidScopeMessage = "scope not allowed by origin client configuration";
+    String subjectToken = request.getRequestParameters().get("subject_token");
+    if (properties.getJwtProfile().isTokenExchangeDisableUpscoping() && !subjectToken.isBlank()) {
+      invalidScopeMessage = "scope not allowed by subject token configuration";
+      try {
+        // this assumes that iam.access_token.include_scope=true
```
enricovianello: in case it's false probably scopes can be retrieved from TokenRequest object. 
We could try to implement a method "extractScopes" that receives the subjectToken and the request as arguments and returns the String[] of scopes

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/exchange/DefaultTokenExchangePdp.java
```diff
@@ -99,14 +109,29 @@ private TokenExchangePdpResult verifyScopes(TokenExchangePolicy p, TokenRequest
...
+      } catch (Exception e) {
+        throw new InvalidRequestException("Error parsing subject token: " + e.getMessage());
+      }
+    }
 
     for (String scope : request.getScope()) {
-      // Check requested scope is permitted by client configuration
-      if (originClientMatchers.stream().noneMatch(m -> m.matches(scope))) {
-        return invalidScope(p, scope, "scope not allowed by origin client configuration");
+      if (!scope.equals("offline_access")
```
enricovianello, line 132: why offline_access scope is excluded?

enricovianello (changes requested):
Some questions and a fix. But it's almost ready I think

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/exchange/DefaultTokenExchangePdp.java
enricovianello: Catching Throwable instead of Exception could be a solution. Also checking the incòude_scope value too.

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/exchange/DefaultTokenExchangePdp.java
```diff
@@ -99,14 +109,29 @@ private TokenExchangePdpResult verifyScopes(TokenExchangePolicy p, TokenRequest
...
     }
 
-    // The requested scopes must be allowed by the origin client (destination is impersonating the
-    // origin client)
-    Set<ScopeMatcher> originClientMatchers = scopeMatcherRegistry.findMatchersForClient(origin);
+    Set<ScopeMatcher> scopeMatchers = scopeMatcherRegistry.findMatchersForClient(origin);
+    String invalidScopeMessage = "scope not allowed by origin client configuration";
+    String subjectToken = request.getRequestParameters().get("subject_token");
+    if (properties.getJwtProfile().isTokenExchangeDisableUpscoping() && !subjectToken.isBlank()
+        && properties.getAccessToken().isIncludeScope()) {
```
enricovianello: Scopes are always stored into database (till now) inside token_scope table. So if scopes are not available within the JWT, you can access the scopes using `IamTokenService`

```
@Autowired
IamTokenService tokenService;
...
OAuth2AccessTokenEntity at = tokenService.readAccessToken(subjectToken);
at.getScopes().foreach(s -> ...)
```

So you can always build the scopeMatcher collection and not rely also on isIncludeScope condition.

enricovianello (changes requested):
In case isIncludeScope is false we can retrieve the scopes from database so we can always create and check the scope matchers

iam-login-service/src/main/resources/application.yml
```diff
@@ -109,6 +109,7 @@ iam:
 
   jwt-profile:
     default-profile: ${IAM_JWT_DEFAULT_PROFILE:iam}
+    token-exchange-disable-upscoping: ${IAM_TOKEN_EXCHANGE_DISABLE_UPSCOPING:false}
```
enricovianello: just a final change (I apologize!)
it's better to have "enabled" configuration parameters
Probably `token-exchange-upscoping-enabled` with default value true is more compliant to all the others inside the application.yml

Sorry! :)

enricovianello (changes requested):
Don't hate me :)

iam-login-service/src/main/java/it/infn/mw/iam/api/common/client/RegisteredClientDTO.java
```diff
@@ -273,6 +273,10 @@ public class RegisteredClientDTO {
       ClientViews.NoSecretDynamicRegistration.class, ClientViews.DynamicRegistration.class})
   private boolean active;
 
+  @JsonView({ClientViews.Limited.class, ClientViews.Full.class, ClientViews.ClientManagement.class,
```
enricovianello: The ClientViews.Full view is not necessary because it extends the Limited one. We can omit it here.

S203932: It has been removed :+1:

enricovianello (approved):

PR #1039, opened 2025-08-27
Add support for voperson_external_affiliation in the AARC JWT profile

mmutadich:


iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/aarc/AarcJWTProfileTokenIntrospectionHelper.java
```diff
@@ -85,13 +94,27 @@ public Map<String, Object> assembleIntrospectionResult(OAuth2AccessTokenEntity a
             claimValueHelper.getClaimValueFromUserInfo(EDUPERSON_ASSURANCE, iamUserInfo));
       }
 
-      result.put("voperson_id", userInfo.getSub());
```
rmiccoli, line 88: Was the `voperson_id` claim removed by mistake?

mmutadich: Yes it was a mistake! I have put it back :)

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/aarc/AarcJWTProfileTokenIntrospectionHelper.java
```diff
@@ -85,13 +94,27 @@ public Map<String, Object> assembleIntrospectionResult(OAuth2AccessTokenEntity a
...
-
+      OAuth2Authentication auth = accessToken.getAuthenticationHolder().getAuthentication();
+      SavedUserAuthentication userAuth = (SavedUserAuthentication) auth.getUserAuthentication();
```
rmiccoli: This check is duplicated here, as it is already present in the `process` method. I'd leave the check there, while here I'd do
```
OAuth2Authentication auth = accessToken.getAuthenticationHolder().getAuthentication();
Map<String, String> claims = extAuthnProcessor.process(auth);

if (!claims.isEmpty()) {
    result.put(VOPERSON_EXTERNAL_AFFILIATION, addVoPersonExternalAffiliation(claims));
}
```
and
```
private String addVoPersonExternalAffiliation(Map<String, String> claims) {
    if (claims.containsKey("EPSA")) {
        return claims.get("EPSA");
    } 
    if (claims.containsKey("eduperson_scoped_affiliation")) {
        return claims.get("eduperson_scoped_affiliation");
    }
    return "";
}
```
But now I have a doubt: do we want to insert an empty string in the `voperson_external_affiliation` claim if neither the _EPSA_ nor the _eduperson_scoped_affiliation_ claims are present in the additional info?
If not, we should do something like
```
OAuth2Authentication auth = accessToken.getAuthenticationHolder().getAuthentication();
Map<String, String> claims = extAuthnProcessor.process(auth);
String affiliation = addVoPersonExternalAffiliation(claims);
if (!affiliation.isEmpty()) {
    result.put(VOPERSON_EXTERNAL_AFFILIATION, affiliation);
}
```

mmutadich: In `AarcJWTProfileUserinfoHelper` I found that if I didn't set the `voperson_external_affiliation` claim I would get a _NullPointerException_ so to avoid this I would set the value to an empty string. I thought it would be the same for the `AarcJWTProfileTokenIntrospectionHelper` so I used the empty string here as well but when I made the changes you recommend it was fine! I have updated the file with these changes but I don't understand why the _NullPointerException_ doesn't occur in the `AarcJWTProfileTokenIntrospectionHelper`.

rmiccoli: Exactly, in the case of user info, when the `voperson_external_affiliation` claim is null, you get _NullPointerException_ [here](https://github.com/indigo-iam/iam/blob/issue-466/iam-login-service/src/main/java/it/infn/mw/iam/core/userinfo/IamUserInfoEndpoint.java#L79) when the UserInfo object is serialised to JSON. Instead, if the `voperson_external_affiliation` claim is empty, it is then discarded (see [here](https://github.com/indigo-iam/iam/blob/issue-466/iam-login-service/src/main/java/it/infn/mw/iam/core/userinfo/UserInfoResponse.java#L99)) and therefore not returned in the response to UserInfo. We could add a check [here](https://github.com/indigo-iam/iam/blob/issue-466/iam-login-service/src/main/java/it/infn/mw/iam/core/userinfo/AarcDecoratedUserInfo.java#L67) (if getVoPersonExternalAffiliation() not null), but I would leave it as it is now.

However, in the case of introspection endpoint, the logic is different: the claim is not inserted in [result](https://github.com/indigo-iam/iam/blob/issue-466/iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/aarc/AarcJWTProfileTokenIntrospectionHelper.java#L54) if it is null.

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/aarc/AarcJWTProfileTokenIntrospectionHelper.java
```diff
@@ -85,13 +93,27 @@ public Map<String, Object> assembleIntrospectionResult(OAuth2AccessTokenEntity a
             claimValueHelper.getClaimValueFromUserInfo(EDUPERSON_ASSURANCE, iamUserInfo));
       }
 
+      OAuth2Authentication auth = accessToken.getAuthenticationHolder().getAuthentication();
+      Map<String, String> claims = extAuthnProcessor.process(auth);
+      String authentication = addVoPersonExternalAffiliation(claims);
+      if (!authentication.isEmpty()){
+        result.put(VOPERSON_EXTERNAL_AFFILIATION, addVoPersonExternalAffiliation(claims));
```
rmiccoli: We can directly put `authentication` in place of `addVoPersonExternalAffiliation(claims)`.

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/aarc/AarcJWTProfileTokenIntrospectionHelper.java
mmutadich: Ah I see, that makes much more sense! Thank you for the explanation :)

enricovianello:
Superseded by this commit https://github.com/indigo-iam/iam/commit/894102165f6ef5926cea6e114b114314d6d66507 
This work was imported and included by that PR.
I'm closing this. Thanks again @mmutadich :)

PR #1046, opened 2025-09-09
Add code to handle of Immediate Group Requests in the Database

### LINKED ISSUE #979: Group request created for parent from if trying to join subgroup but getting accepted to subgroup automatically grant membership to parent group.
Hi INDIGO IAM developers.

The group joining request is quite confusing for IAM. If you join a subgroup a group joining request will be created for all the parent group as currently subgroup membership implies a membership of parent group. 

The issue is that the if you are accepted into the subgroup, but the request to join parent group will remain. 

**Preferred solution**
- filter so that only the (leaf group) request will create a request and no request will be made for parent group
- Auto removal of request after the subgroup request is accepted/rejected.
I edited out the preferred, I don't have a personal preference for either one of them

Sae126V:
Resolves https://github.com/indigo-iam/iam/issues/979 

say, we have rootGroup, subGroup1, subGroup2, subGroup3.

If the user raises a request to join subGroup3. We will auto-approve requests to the parent group requests if the subGroup3 Group manager approves the request.

### DIFF
--- a/iam-persistence/src/main/java/it/infn/mw/iam/persistence/repository/IamGroupRequestRepository.java
+++ b/iam-persistence/src/main/java/it/infn/mw/iam/persistence/repository/IamGroupRequestRepository.java
@@ -35,5 +35,5 @@ public interface IamGroupRequestRepository
   List<IamGroupRequest> findByUsernameAndGroup(@Param("username") String username,
       @Param("groupName") String groupName);
 
-      Optional<IamGroupRequest> findByGroupIdAndAccountIdAndStatus(Long groupID, Long accountID, IamGroupRequestStatus status);
+  Optional<IamGroupRequest> findByGroupIdAndAccountIdAndStatus(Long groupID, Long accountID, IamGroupRequestStatus status);
 }
--- a/iam-persistence/src/main/java/it/infn/mw/iam/persistence/repository/IamGroupRequestRepository.java
+++ b/iam-persistence/src/main/java/it/infn/mw/iam/persistence/repository/IamGroupRequestRepository.java
@@ -23,6 +23,7 @@ import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.PagingAndSortingRepository;
 import org.springframework.data.repository.query.Param;
 
+import it.infn.mw.iam.core.IamGroupRequestStatus;
 import it.infn.mw.iam.persistence.model.IamGroupRequest;
 
 public interface IamGroupRequestRepository
@@ -34,5 +35,5 @@ public interface IamGroupRequestRepository
   List<IamGroupRequest> findByUsernameAndGroup(@Param("username") String username,
       @Param("groupName") String groupName);
 
-  
+      Optional<IamGroupRequest> findByGroupIdAndAccountIdAndStatus(Long groupID, Long accountID, IamGroupRequestStatus status);
 }
--- a/iam-login-service/src/main/java/it/infn/mw/iam/api/requests/service/DefaultGroupRequestsService.java
+++ b/iam-login-service/src/main/java/it/infn/mw/iam/api/requests/service/DefaultGroupRequestsService.java
@@ -19,6 +19,8 @@ import static it.infn.mw.iam.core.IamGroupRequestStatus.APPROVED;
 import static it.infn.mw.iam.core.IamGroupRequestStatus.PENDING;
 import static it.infn.mw.iam.core.IamGroupRequestStatus.REJECTED;
 
+import static java.util.Objects.isNull;
+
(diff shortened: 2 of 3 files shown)


giacomini:
Let me see if I understand the solution: given the hierarcy rootGroup, subGroup1, subGroup2, subGroup3, if a user requests to join subGroup3 and the request is approved by the subGroup3 group manager, the user is automatically added to subGroup2, subGroup1 and rootGroup. Is my understanding correct?

rmiccoli:
Hi @Sae126V, I don't understand what changes from the current behaviour. See https://github.com/indigo-iam/iam/issues/979#issuecomment-3280372101

Sae126V:
> Let me see if I understand the solution: given the hierarcy rootGroup, subGroup1, subGroup2, subGroup3, if a user requests to join subGroup3 and the request is approved by the subGroup3 group manager, the user is automatically added to subGroup2, subGroup1 and rootGroup. Is my understanding correct?

Yep. That is correct. The fix is now just for the requests waiting in the table - we are auto approving the requests with PENDING state to approved state for all the rootGroup, subGroup1, subGroup2, if subGroup3 group manager approves the user request.

Sae126V:
> Hi @Sae126V, I don't understand what changes from the current behaviour. See [#979 (comment)](https://github.com/indigo-iam/iam/issues/979#issuecomment-3280372101)

The fix is for the "requests" waiting to be approved or rejected based on group manager action i.e, We will auto-approve "requests" to the parent group requests if the subGroup3 Group manager approves the request. The behaviour is the same as before, the fix is for the "requests" only.

rmiccoli:
We demonstrated the issue in this way:
- Admin creates a rootGroup and then subGroup1 (rootGroup/subGroup1), subGroup2  (rootGroup/subGroup1/subGroup2) and subGroup3 (rootGroup/subGroup1/subGroup2/subGroup3)
- A user requests to join subGroup1 and subGroup3
- If the admin or group manager approves the request to join subGroup3, the user is automatically added to subGroup2 and subGroup1 as well, but the request for subGroup1 remains PENDING
  - this PR fixes the issue by changing the status of the parent group request to APPROVED
- If the admin or group manager rejects the request for subGroup3, the user is not added to that group, but the request to join subGroup1 remains (correct behaviour)

A question about current behaviour:
- if the admin or group manager rejects the request for subGroup1, the user is not added to that subgroup but the request for subGroup3 remains. If the admin or manager of subGroup3 approves the request, the user is automatically added to subGroup2 and subGroup1 as well. In this case, we believe it's better to reject all requests related to subgroups.

Considering that automatic approval for parent groups needs to be reconsidered, what do you think is the best solution for this use case at the moment? @giacomini @enricovianello

giacomini:
> A question about current behaviour:
> 
>     * if the admin or group manager rejects the request for subGroup1, the user is not added to that subgroup but the request for subGroup3 remains. If the admin or manager of subGroup3 approves the request, the user is automatically added to subGroup2 and subGroup1 as well. In this case, we believe it's better to reject all requests related to subgroups.
> 
> 
> Considering that automatic approval for parent groups needs to be reconsidered, what do you think is the best solution for this use case at the moment? @giacomini @enricovianello

I agree that with the current behavior (which we need to reconsider) what you suggest is the most sensible approach.

F.

iam-persistence/src/main/java/it/infn/mw/iam/persistence/repository/IamGroupRequestRepository.java
```diff
@@ -34,5 +35,5 @@ public interface IamGroupRequestRepository
   List<IamGroupRequest> findByUsernameAndGroup(@Param("username") String username,
       @Param("groupName") String groupName);
 
-  
+      Optional<IamGroupRequest> findByGroupIdAndAccountIdAndStatus(Long groupID, Long accountID, IamGroupRequestStatus status);
```
enricovianello: Indentation to be fixed (?)

enricovianello (approved):

enricovianello (changes requested):
Just a small indentation fix (probably)

iam-persistence/src/main/java/it/infn/mw/iam/persistence/repository/IamGroupRequestRepository.java
Sae126V: Ahh! I missed it, Sorry. Fixed it now in [c0d6d0a](https://github.com/indigo-iam/iam/pull/1046/commits/c0d6d0a22e963d10a9cb5e2475db696e6d46da83)

enricovianello (approved):

PR #1048, opened 2025-09-15
Add a table to persist the admin key and to perform re-encryption

Sae126V:
This is to help admin rotate the KEY used for encrypting and decrypting the secrets, if required. As of now there is NO logic to force the admins to rotate the keys - We can do that(Please let me know your thoughts on it, team).

> [!important]
This PR changes the default value for `mfa.password-to-encrypt-and-decrypt` from `define_me_please` to *undefined*. In case you were using this default value, you can use the logic introduced by this PR in order to rotate the secret to a better value.


iam-persistence/src/main/resources/db/migration/h2/V116__store_mfa_admin_key.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE iam_totp_admin_key (
```
enricovianello, line 1: Having reached V113 on iam-persistence after v1.13.0, this file needs to be renamed to V114

iam-persistence/src/main/resources/db/migration/mysql/V116__store_mfa_admin_key.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE iam_totp_admin_key (
```
enricovianello, line 1: Same renaming said above required

iam-login-service/src/main/java/it/infn/mw/iam/core/mfa/IamTotpMfaKeyRotationRunner.java
```diff
@@ -0,0 +1,106 @@
...
+import it.infn.mw.iam.util.mfa.IamTotpMfaInvalidArgumentError;
+
+/**
+ * Runs at application startup (when the "mfa" profile is active) and ensures the
+ * configured MFA admin key is recorded; if the key changed, it re-encrypts all
+ * stored TOTP secrets using the old key -> And, encrypts it using new key.
+ */
+@Component
+@Profile("mfa")
+public class IamTotpMfaKeyRotationRunner implements ApplicationRunner {
```
enricovianello, line 32: nice this! I was not aware of it! Good catch 👍

iam-login-service/src/main/java/it/infn/mw/iam/core/mfa/IamTotpMfaKeyRotationRunner.java
```diff
@@ -0,0 +1,106 @@
...
+    this.iamTotpSecretRotationService = iamTotpSecretRotationService;
+  }
+
+  @Override
+  public void run(ApplicationArguments args) {
+    /**
+     * The current admin key used to encrypt/decrypt TOTP secrets.
+     * MUST be provided when MFA is enabled.
+     */
+    final String currentMfaKey = iamTotpMfaProperties.getPasswordToEncryptOrDecrypt();
```
enricovianello: You could evaluate to use @Value on a class field for these two secrets

e.g.

```
    @Value("${mfa.password-to-encrypt-and-decrypt}")
    private String passwordToEncryptAndDecrypt;

    @Value("${mfa.old-password-to-decrypt:}")
    private String oldPasswordToDecrypt;
```

iam-login-service/src/main/java/it/infn/mw/iam/core/mfa/IamTotpMfaKeyRotationRunner.java
```diff
@@ -0,0 +1,106 @@
...
+     * Optional previous admin key; required only when rotating to a new key so that
+     * existing secrets can be re-encrypted.
+     */
+    final String previousMfaKey = iamTotpMfaProperties.getOldPasswordToDecrypt();
+
+    final String storedHash = repository.findCurrentHash();
+
+    // First-time setup: require a key and store its hash
+    if (storedHash == null) {
+      if (!StringUtils.hasText(currentMfaKey)) {
```
enricovianello: by default you have "define_me_please" so this case could be probably not a first setup but a "wrong setup"

enricovianello: I would prefer to have not the default "define_me_please" value in order to fail the startup if a value is not provided.

iam-login-service/src/main/resources/application-mfa.yml
```diff
@@ -16,4 +16,8 @@
 
 mfa:
   multi-factor-settings-btn-enabled: ${IAM_TOTP_MFA_ENABLE_MFA_SETTINGS_BUTTON:true}
+  # Please store this KEY securly & safetly
```
enricovianello, line 19: The default value of `password-to-encrypt-and-decrypt` is not a good choice. But this is not your fault :) Luckily MFA is not yet really used by IAM deployers but if this value is not changed when enabling mfa profile, it's used! I'd prefer an approach of (fail if you add mfa profile and you don't set this). We need to discuss this and probably write good release notes if we're going to change this default!

enricovianello: Your rotation mechanism will solve this problem for sure!

enricovianello (changes requested):
Change migration version is the only thing I found. In general it seems a necessary feature! Having set "define_me_please" as a default key value that probably is currently used by all the current INDIGO IAM instances with MFA enabled! Doh!

iam-persistence/src/main/resources/db/migration/h2/V116__store_mfa_admin_key.sql
Sae126V: Done. Updated the name

iam-persistence/src/main/resources/db/migration/mysql/V116__store_mfa_admin_key.sql
Sae126V: Done. Updated the name

iam-login-service/src/main/java/it/infn/mw/iam/core/mfa/IamTotpMfaKeyRotationRunner.java
Sae126V: Cool. Updated the code. Done

iam-login-service/src/main/java/it/infn/mw/iam/core/mfa/IamTotpMfaKeyRotationRunner.java
Sae126V: Make sense to NOT have a default value.  Done

iam-login-service/src/main/resources/application-mfa.yml
Sae126V: Cool. Updated the code now.

iam-login-service/src/main/java/it/infn/mw/iam/config/mfa/IamTotpMfaProperties.java
```diff
@@ -24,6 +24,7 @@ public class IamTotpMfaProperties {
 
   private boolean multiFactorSettingsBtnEnabled;
   private String passwordToEncryptAndDecrypt;
+  private String oldPasswordToDecrypt;
 
   public String getPasswordToEncryptOrDecrypt() {
```
rmiccoli: Small typo here: getPasswordToEncryptAndDecrypt

Sae126V: My bad to not change the method name. Done.

enricovianello (approved):

PR #1049, opened 2025-09-15
Implement Explicit Client Registration

### LINKED ISSUE #1004: Implement Explicit Client Registration with Entity Statement (OP Role)
**Description**
Implement support for OpenID Federation Explicit Client Registration in the OpenID Provider (OP) role as specified in [OpenID Connect Federation 1.0, Section 12.2](https://openid.net/specs/openid-federation-1_0.html#name-explicit-registration), focusing on Entity Statement submission. The feature enables Relying Parties (RPs) to dynamically register with an OpenID Provider using a request with content type `application/entity-statement+jwt`.

**Note**
Trust Chain submission (`application/trust-chain+json`) can be considered at a later stage.

rmiccoli:
### Implementation details
* Created a new federation registration endpoint (e.g. `https://<iam>/iam/api/oid-fed/client-registration`)
* Added an expiry date to the registered client that is the same as the expiry date of the resolved trust chain
* Added a task running _once a day_ to check if there are expired clients
   * the expired clients are disabled

Bound to https://github.com/indigo-iam/OpenID-Connect-Java-Spring-Server/pull/30


iam-persistence/src/main/resources/db/migration/mysql/V110__client_federation_metadata.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE IF NOT EXISTS client_federation_metadata (
+  client_details_id BIGINT PRIMARY KEY,
+  expiration TIMESTAMP NOT NULL,
+  entity_id VARCHAR(512) NOT NULL UNIQUE);
```
enricovianello: UNIQUE makes sense but for future scenarios where we could imagine to map a vo to a root group maybe this is a useless constraint, knowing that we're checking inside the codebase if the RP entity id is present or not

iam-persistence/src/main/resources/db/migration/mysql/V110__client_federation_metadata.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE IF NOT EXISTS client_federation_metadata (
```
enricovianello: I'd suggest something like `client_relying_party` to clarify that the entity id is related to the external relying party

iam-login-service/src/main/java/it/infn/mw/iam/api/client/service/ClientConverter.java
```diff
@@ -126,9 +126,12 @@ public RegisteredClientDTO registeredClientDtoFromEntity(ClientDetailsEntity ent
...
 
-    clientDTO.setCreatedAt(entity.getCreatedAt());
-    if (entity.getClientLastUsed() != null) {
-      clientDTO.setLastUsed(entity.getClientLastUsed().getLastUsed());
+    clientDTO.setCreatedAt(entity.getCreatedAt());
+    if (entity.getClientLastUsed() != null) {
+      clientDTO.setLastUsed(entity.getClientLastUsed().getLastUsed());
+    }
+    if (entity.getFederationMetadata() != null) {
+      clientDTO.setExpiration(entity.getFederationMetadata().getExpiration());
```
enricovianello: add also the entityID

iam-login-service/src/test/java/it/infn/mw/iam/test/service/client/ClientRegistrationServiceTests.java
```diff
@@ -301,15 +301,14 @@ void testAllowedGrantTypeChecks() throws ParseException {
...
-          service.registerClient(request, noAuth);
-        });
+    exception = Assertions.assertThrows(InvalidClientRegistrationRequest.class, () -> {
+      RegisteredClientDTO request = new RegisteredClientDTO();
+      request.setClientName("example");
+      request.setGrantTypes(
+          Sets.newHashSet(AuthorizationGrantType.CODE, AuthorizationGrantType.TOKEN_EXCHANGE));
+      request.setRedirectUris(Sets.newHashSet("https://example/cb"));
+      service.registerClient(request, noAuth);
+    });
```
enricovianello, line 311: Probably we need to check that a client registered through the explicit reg. endpoint can be edited through the management API endpoints without losing the Relying Party one-to-one related

iam-login-service/src/main/java/it/infn/mw/iam/api/client/registration/service/DefaultClientRegistrationService.java
```diff
@@ -359,20 +360,26 @@ public RegisteredClientDTO registerClient(RegisteredClientDTO request,
     client.setDynamicallyRegistered(true);
     client.setActive(true);
 
+    if (request.getEntityId() != null) {
```
enricovianello: use a method called e.g. hasRelyingParty() that returns a boolean to improve readability

iam-login-service/src/main/java/it/infn/mw/iam/api/openid_federation/FederationRegistrationController.java
```diff
@@ -0,0 +1,162 @@
...
+  public ResponseEntity<String> register(@RequestBody String requestJwt)
+      throws ParseException, BadJOSEException, JOSEException, java.text.ParseException {
+
+    // 1. Parse request Entity Statement (self-signed EC of the RP)
+    EntityStatement rpRequest = EntityStatement.parse(requestJwt);
+
+    Optional<ClientDetailsEntity> existingClient =
+        clientRepo.findByEntityId(rpRequest.getEntityID().getValue());
+    if (existingClient.isPresent()) {
+      clientRepo.delete(existingClient.get());
```
enricovianello: Move this later, after point 6 and use ClientService instead so the delete is added to AUDIT log

iam-login-service/src/main/java/it/infn/mw/iam/api/openid_federation/FederationRegistrationController.java
```diff
@@ -0,0 +1,162 @@
...
+    RegisteredClientDTO registeredClient =
+        clientRegistrationService.registerClient(dtoClient, null);
+
+    // 6. Build the response (Entity Statement)
+    String responseEs = federationResponseBuilder.build(registeredClient, trustChain);
+
+    return ResponseEntity.ok()
+      .contentType(MediaType.valueOf("application/explicit-registration-response+jwt"))
+      .body(responseEs);
+  }
```
enricovianello, line 197: Add Exception handler

iam-login-service/src/main/java/it/infn/mw/iam/api/openid_federation/FederationRegistrationController.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    // 3. Resolve and validate trust chain starting from received EC
+    TrustChain trustChain = trustChainService.validateFromEntityConfiguration(rpRequest);
+
+    // 4. Create RegisteredClientDTO from RP metadata
+    RegisteredClientDTO dtoClient = createClientDtoFromRpMetadata(rpRequest);
+    Date clientExpiration = trustChain.resolveExpirationTime();
+    dtoClient
+      .setExpiration(LocalDate.ofInstant(clientExpiration.toInstant(), ZoneId.systemDefault()));
```
enricovianello: Change this type on the related MitreID PR: from LocalDate to Date

enricovianello (changes requested):

iam-login-service/src/test/java/it/infn/mw/iam/test/service/client/ClientRegistrationServiceTests.java
rmiccoli: From spec (section [12.2.2](https://openid.net/specs/openid-federation-1_0.html#section-12.2.2)), I understand that a new explicit registration request is needed to update a client registered through federation registration endpoint.

_"The OP SHOULD NOT provision the RP with a registration_access_token and a registration_client_uri because the expected way for the RP to update its registration is to make a new Explicit Registration request. If the RP is provisioned with a registration_access_token for some purpose, for example to let it independently check its registered metadata, the token MUST NOT allow modification of the registration."_

enricovianello (approved):

PR #1056, opened 2025-09-26
Tidy up role-reader view

### LINKED ISSUE #1042: Wrong information displayed when viewing with other accounts with READER_ROLE
With reference to issue [794](https://github.com/indigo-iam/iam/issues/794)/pull request [842](https://github.com/indigo-iam/iam/pull/842), the `ROLE_READER` has been introduced. 

When assigned to a user, it enables the user to be able to view additional information, here among which other user accounts. 

However, when viewing other accounts, the group requests displayed are not related to the account being viewed, but to the account with the monitoring role. This can be seen from the image below. 

Image

From the image, one can see  2 group requests. These are from the "Patrick's Test" user, not the "Test User" currently being viewed. Moreover, one can submit new group requests or remove previous group requests, but these requests does not pertain to the user being viewed, but the user doing the viewing. 

This is the information related to the user with the monitoring role, "Patrick's Test" user:

Image

These are the group requests (as viewed from an Admin account, in this case the "Patrick Machalet" user). 

Image

Also, given that one has the `ROLE_READER`, then one has the ability to reset the password of other accounts. Not sure if this is intentional?

Lastly, whilst monitoring another account one is displayed the option of changing the given users details (username and profile picture), but one is not actually allowed to do this, as seen from the image below. 

Image

garaimanoj:


rmiccoli (changes requested):
Hi @garaimanoj, I noticed that users with `reader` role can delete other users' group requests, even if the backend does not allow it (_access denied_). Therefore, the only thing missing is to hide the delete button for the request.

Also, now the group manager can also see requests to join groups that they do not manage (as if they were an admin), although an error occurs if they try to approve or reject them.

garaimanoj:
> Hi @garaimanoj, I noticed that users with `reader` role can delete other users' group requests, even if the backend does not allow it (_access denied_). Therefore, the only thing missing is to hide the delete button for the request.
> 
> Also, now the group manager can also see requests to join groups that they do not manage (as if they were an admin), although an error occurs if they try to approve or reject them.

Thanks for your review, I have now updated the code to hide those button.

rmiccoli (approved):

iam-login-service/src/main/java/it/infn/mw/iam/api/requests/GroupRequestUtils.java
```diff
@@ -124,7 +125,9 @@ public Set<String> getManagedGroups() {
   }
 
   public boolean isPrivilegedUser() {
```
enricovianello: Not a problem for this PR but a comment/to-do for the future:

I'd move this method inside accountUtils. It's used only by DefaultGroupRequestService.
Other cleaner alternatives that make code more readable are:
- define boolean hasAuthorities(IamAuthority a1, ...) or (String a1, ...) method inside account utils that checks the current authenticated user has the list of authorities
- inside DefaultGroupRequestService calling accountUtils.hasAuthorities("ROLE_ADMIN", "ROLE_READER")
so this method could be used also somewhere

iam-login-service/src/main/webapp/resources/iam/apps/dashboard-app/components/group/members/group.members.component.html
```diff
@@ -22,16 +22,16 @@
 
 		<div class="row iam-div-row"
```
enricovianello, line 23: I know that we're putting effort on the new dashboard but these changes should not be so difficult to add:

I would add a div visibile for non privileged users (not admin not group manager not reader) with a "Not enough privileges" message.
Then I'd move the ng-if=" is admin or is reader or is group manager" inside the other div and a finer grained authz for buttons (only for admins and group managers) for the delete button for example

Does this make sense?

enricovianello (commented):
Just a few comments or changes. If there's no time to fix them soon we can proceed with the merge

iam-login-service/src/main/java/it/infn/mw/iam/api/requests/GroupRequestUtils.java
garaimanoj: I have moved the method to account utils so that it can be reused. 

While working I have noticed that `Requests` page/tab is not visible to `Reader` although reader can see the group request from user details page, so I have made `Requests` page visible to `Reader` only for group requests. If the reader is also a group manager then the user can take action on the requests. With only `Reader` role group requests are view only.

image

iam-login-service/src/main/webapp/resources/iam/apps/dashboard-app/components/group/members/group.members.component.html
garaimanoj: Non privileged uses do not have access to group details or members of the group area.

iam-login-service/src/main/java/it/infn/mw/iam/api/requests/GroupRequestUtils.java
enricovianello: Top!

iam-login-service/src/main/webapp/resources/iam/apps/dashboard-app/components/group/members/group.members.component.html
enricovianello: Reading again the code changes and this time with less fever I agree it's ok as it is :)

enricovianello (approved):

PR #1065, opened 2025-10-03
Add openid_relying_party metadata

### LINKED ISSUE #1060: Add support for `openid_relying_party` metadata

rmiccoli:


enricovianello (approved):
Next time i'd like to define an interface with all the metadata names as final String, in order to be compliant with static analysis and not write constants here. But it's ok the PR.

PR #1069, opened 2025-10-24
Implement automatic client registration

### LINKED ISSUE #1059: Implement Automatic Client Registration (OP Role)
Implement OpenID Federation Automatic Client Registration (OP Role)

rmiccoli:
This PR implements one of the client/Relying Party registration mechanisms defined by OpenID Federation.

Automatic Registration allows an RP to send Authentication Requests without a prior registration step with the OP. When the RP reaches the OP’s Authorization Endpoint, the OP automatically registers the RP if it can build and validate a trust chain for the RP - either by resolving the RP’s Entity Configuration from its `entity_id` or by using a provided `trust_chain` - and a common Trust Anchor exists.

At the Authorization Endpoint, in fact, the RP can use a Request Object by sending the `request` parameter, a signed JWT containing some required claims (`aud`, `client_id`, `iss`, `jti`) and optionally `iat` and `trust_chain`.

Note that the OP assigns the `client_id` to the RP, which must correspond to its `entity_id` (URL).

Reference: https://openid.net/specs/openid-federation-1_0.html#name-automatic-registration


federicaagostini:
Remember also to run the [disableExpiredClients](https://github.com/indigo-iam/iam/blob/develop/iam-login-service/src/main/java/it/infn/mw/iam/config/TaskConfig.java#L152) task only when `openid-fed` is included among the active profiles.

rmiccoli:
> Remember also to run the [disableExpiredClients](https://github.com/indigo-iam/iam/blob/develop/iam-login-service/src/main/java/it/infn/mw/iam/config/TaskConfig.java#L152) task only when `openid-fed` is included among the active profiles.

Done :)

iam-login-service/src/main/java/it/infn/mw/iam/config/BasicAuthLoggingFilter.java
```diff
@@ -0,0 +1,65 @@
...
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+import org.springframework.web.filter.OncePerRequestFilter;
+
+@Component
```
enricovianello, line 31: Add `@Profile('openid-federation')`

rmiccoli: Done

iam-login-service/src/main/java/it/infn/mw/iam/core/oidc/IamAuthorizationRequestFilter.java
```diff
@@ -0,0 +1,479 @@
...
+
+  private final Environment env;
+  private final IamClientRepository clientRepo;
+  private final DefaultClientManagementService clientManagementService;
+  private final RedirectResolver redirectResolver;
+  private final TrustChainService trustChainService;
+  private final AutomaticClientRegistrationMapper clientMapper;
+
+  private LoginHintExtracter loginHintExtracter = new RemoveLoginHintsWithHTTP();
+  private RequestMatcher requestMatcher = new AntPathRequestMatcher("/authorize");
```
enricovianello: I'd suggest to convert it into a static final constant and to rename it into something like "authorizeMatcher"

iam-login-service/src/main/java/it/infn/mw/iam/core/oidc/TrustChainService.java
```diff
@@ -26,7 +25,6 @@
 import com.nimbusds.openid.connect.sdk.federation.trust.TrustChain;
 
 @Service
-@Profile("openid-federation")
```
enricovianello, line 29: This means we have always an instance of a TrustChain service. Even if `openid-federation` is not specified within the active profiles. Why? It is needed by others?

iam-login-service/src/main/java/it/infn/mw/iam/core/oidc/TrustChainResolver.java
enricovianello, line 42: Should it be added a `@Profile("openid-federation")` ? This resolver could not exist in the other cases

iam-login-service/src/main/java/it/infn/mw/iam/config/MitreServicesConfig.java
```diff
@@ -187,21 +185,6 @@ ServerConfigInterceptor serverConfigInterceptor() {
...
-  FilterRegistrationBean<AuthorizationRequestFilter> disabledMitreFilterRegistration(
-      AuthorizationRequestFilter f) {
-
-    FilterRegistrationBean<AuthorizationRequestFilter> b = new FilterRegistrationBean<>(f);
-    b.setEnabled(false);
-    return b;
-  }
-
-  @Bean(name = "mitreAuthzRequestFilter")
-  AuthorizationRequestFilter authorizationRequestFilter() {
```
enricovianello, line 200: In order to not add openid fed logic when no profile is added, I tend to keep the MitreID implementation when no openid-federation" profile is present.
Otherwise we can have this implementation.

Importing mitreID, refactor it and make something like fed one extends default one could be also done but it's a bigger change.

So something like:

```
  @Bean
  @Profile("!openid-federation")
  AuthorizationRequestFilter authorizationRequestFilter() {
     return new AuthorizationRequestFilter();
  }

  @Bean
  @Profile("openid-federation")
  AuthorizationRequestFilter authorizationRequestFilter(Environment env, IamClientRepository clientRepo, DefaultClientManagementService clientManagementService, RedirectResolver redirectResolver, TrustChainService trustChainService, AutomaticClientRegistrationMapper clientMapper) {
     return new IamAuthorizationRequestFilter(env, clientRepo, clientManagementService, redirectResolver, trustChainService, clientMapper);
  }
```

Evaluate to move the second bean inside IamConfig. But probably they should be kept one aside the other. So Mitre services can be a tolerated place XD
I don't think qualifiers are required.
This means also to remove the exclusion of AuthorizationRequestFilter in the main class.

iam-login-service/src/main/java/it/infn/mw/iam/core/oidc/IamAuthorizationRequestFilter.java
```diff
@@ -0,0 +1,479 @@
...
+      chain.doFilter(req, res);
+      return;
+    }
+
+    Map<String, String> params = createRequestMap(request.getParameterMap());
+
+    Optional<ClientDetailsEntity> client = Optional.empty();
+
+    String clientId = params.get(CLIENT_ID);
+    if (clientId != null) {
```
enricovianello: no else for clientId null, probably a sendAuthenticationError could be necessary.

enricovianello (changes requested):

enricovianello (approved):

PR #1076, opened 2025-11-03
Add option to save access tokens on database or not

### LINKED ISSUE #537: Option needed for IAM not to store access tokens in the DB

enricovianello:


enricovianello:
Replaced by #1165

PR #1093, opened 2025-11-20
Evaluate all SAML id resolvers

### LINKED ISSUE #1055: User not found after changing IAM_SAML_ID_RESOLVERS order
Hi,

IMO, current processing of `IAM_SAML_ID_RESOLVERS` is broken. Each attribute in the ID resolver list is tried in order and for the first attribute present in the SAML response from the IdP, it looks for a matching user in INDIGO IAM and if none is found, it presents the registration form to the user to create a new account (if enabled). As a result, if a user exists in INDIGO IAM matched with a certain attribute, if later `IAM_SAML_ID_RESOLVERS` is changed to support another attribute with a higher priority (so prepended to the existing list), the existing user is not found anymore (and luckily cannot create a new account as the email is already used). 

For example, let say that I was matched initially with `eduPersonTargetedID` and that at this time the INDIGO IAM instance was using the default `IAM_SAML_ID_RESOLVERS`. Later, `IAM_SAML_ID_RESOLVERS` list was prepended with `subjectID` as part of the effort of promoting the use of GDPR-compliant attributes. So that new users use this new attribute rather than the obsolete one, it was added first in the list. But because of this, next time that I authenticate against INDIGO IAM, as my IdP is providing `subjectID`, I am not recognized anymore: my account is linked to the `eduPersonTargetedID` attribute but the match stopped after failing with `subjectID` instead of trying the next attribute in the list, checking for a possible match.

I think it must really be fixed. I don't see any possible breach as the existing INDIGO IAM user is a legitimate one. If somebody wants to prevent matching with `eduPersonTargetedID`, it must just remove it from `IAM_SAML_ID_RESOLVERS` instead of keeping it at the end of the list...

Note that a mechanism similar to `IAM_SAML_ID_RESOLVERS` was planned (not sure the status) for OIDC and the same approach should be implemented.

rmiccoli:
When the user authenticates to IAM via a SAML IdP, all `IAM_SAML_ID_RESOLVERS` are evaluated and collected to verify whether the user has already been registered and mapped to one of those resolvers. If not, the user is registered by selecting the first resolver found.

### DIFF
--- a/iam-login-service/src/main/java/it/infn/mw/iam/authn/DefaultExternalAccountLinker.java
+++ b/iam-login-service/src/main/java/it/infn/mw/iam/authn/DefaultExternalAccountLinker.java
@@ -81,7 +81,9 @@ public class DefaultExternalAccountLinker implements ExternalAccountLinker {
         (SAMLCredential) token.getExternalAuthentication().getCredentials();
 
     final IamSamlId iamSamlId = samlUserIdResolver.resolveSamlUserIdentifier(credential)
-      .getResolvedId()
+      .getResolvedIds()
+      .stream()
+      .findFirst()
       .orElseThrow(() -> new UsernameNotFoundException(
           "Could not extract a user identifier from the SAML assertion"));
 
@@ -110,5 +112,4 @@ public class DefaultExternalAccountLinker implements ExternalAccountLinker {
     repo.save(targetAccount);
 
   }
-
 }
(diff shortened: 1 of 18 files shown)


enricovianello (changes requested):
The only thing is missing is a test of the logic inside DefaultSAMLUserDetailsService using Mocked args for its constructor that trues to simulate the change in the order of the resolvers.

rmiccoli:
> The only thing is missing is a test of the logic inside DefaultSAMLUserDetailsService using Mocked args for its constructor that trues to simulate the change in the order of the resolvers.

Done :)

enricovianello (approved):

PR #1096, opened 2025-11-24
Bug fix, field comparison is case sensitive

### LINKED ISSUE #1094: Field property settings for notes in the registration form (Mandatory / Optional / Hidden) do not work in v1.13.1
According to the documentation and discussions in issue #759, the field-property configuration should support three modes: Mandatory, Optional, and Hidden.
However, in v1.13.1, changing the field-property value appears to have no effect. Fields remain unchanged regardless of which setting is applied.

```
notes:
  read-only: false
  field-behaviour: hidden
```

S203932:
Fixed the notes field and the affiliation field within the registration page. 
All fields reverted to the default values, as the `populateFieldsWithAdminPreference` function within `iam-login-service/src/main/webapp/resources/iam/apps/registration/registration.controller.js` failed. 
It failed as the assignment is case sensitive. 
Hence, the solution was `$scope.fields[field.toLowerCase()].showField = true;`

### DIFF
--- a/iam-login-service/src/main/webapp/resources/iam/apps/registration/registration.controller.js
+++ b/iam-login-service/src/main/webapp/resources/iam/apps/registration/registration.controller.js
@@ -283,16 +283,16 @@ function RegistrationController(
 				if (
 					$scope.config.fields[field]["fieldBehaviour"].toLowerCase() == "mandatory"
 				) {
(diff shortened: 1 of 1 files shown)


Sae126V:
LGTM! :+1:

enricovianello (approved):

PR #1108, opened 2025-12-02
Don't fallback on default JWT profile

enricovianello:
When the client is allowed to request multiple JWT profiles but the requested scopes clarifies which is the desired profile, don't fallback on IAM profile

### DIFF
--- a/iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/ScopeAwareProfileResolver.java
+++ b/iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/ScopeAwareProfileResolver.java
@@ -35,18 +35,7 @@ public class ScopeAwareProfileResolver implements JWTProfileResolver {
   @Override
   public JWTProfile resolveProfile(Set<String> scopes) {
 
-    if (Objects.isNull(scopes)) {
-      throw new IllegalArgumentException("null list of scopes");
-    }
-    if (scopes.isEmpty()) {
-      return defaultProfile;
-    }
-
-    Set<JWTProfile> matchedProfiles = matches(scopes);
-    if (matchedProfiles.isEmpty() || matchedProfiles.size() > 1) {
-      return defaultProfile;
-    }
-    return matchedProfiles.iterator().next();
+    return resolveProfile(scopes, Set.of());
   }
 
   @Override
@@ -60,14 +49,18 @@ public class ScopeAwareProfileResolver implements JWTProfileResolver {
     }
 
     Set<JWTProfile> clientMatches = matches(clientScopes);
-    if (clientMatches.isEmpty() || clientMatches.size() > 1) {
-      Set<JWTProfile> requestedMatches = matches(requestedScopes);
-      if (requestedMatches.isEmpty() || requestedMatches.size() > 1) {
-        return defaultProfile;
-      }
-      return requestedMatches.iterator().next();
+    if (clientMatches.isEmpty()) {
+      return defaultProfile;
     }
-    return clientMatches.iterator().next();
+    if (clientMatches.size() == 1) {
+      return clientMatches.iterator().next();
+    }
+    // clientMatches.size() > 1
+    Set<JWTProfile> requestedMatches = matches(requestedScopes);
+    if (requestedMatches.isEmpty() || requestedMatches.size() > 1) {
+      return defaultProfile;
+    }
+    return requestedMatches.iterator().next();
   }
 
   private Set<JWTProfile> matches(Set<String> clientScopes) {
--- a/iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/ScopeAwareProfileResolver.java
+++ b/iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/ScopeAwareProfileResolver.java
@@ -19,6 +19,7 @@ import static java.util.stream.Collectors.toCollection;
 
 import java.util.LinkedHashSet;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 public class ScopeAwareProfileResolver implements JWTProfileResolver {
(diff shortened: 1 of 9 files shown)


iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/ScopeAwareProfileResolver.java
```diff
@@ -32,22 +33,48 @@ public ScopeAwareProfileResolver(JWTProfile defaultProfile, Map<String, JWTProfi
   }
 
   @Override
-  public JWTProfile resolveProfile(Set<String> clientScopes) {
+  public JWTProfile resolveProfile(Set<String> scopes) {
```
giacomini, line 39: wouldn't it be better to implement this in terms of the other `resolveProfile`? i.e. something like:
```
public JWTProfile resolveProfile(Set<String> scopes) {
  return resolveProfile(scopes, emptyScopes);
}
```

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/ScopeAwareProfileResolver.java
```diff
@@ -32,22 +33,48 @@ public ScopeAwareProfileResolver(JWTProfile defaultProfile, Map<String, JWTProfi
...
+
+    if (Objects.isNull(clientScopes) || Objects.isNull(requestedScopes)) {
+      throw new IllegalArgumentException("null list of scopes");
+    }
+    if (clientScopes.isEmpty() && requestedScopes.isEmpty()) {
+      return defaultProfile;
+    }
+
+    Set<JWTProfile> clientMatches = matches(clientScopes);
+    if (clientMatches.isEmpty() || clientMatches.size() > 1) {
```
giacomini: isn't this `if (clientMatches.size() != 1) {`?

shouldn't we check that the intersection between client scopes and request scopes, once filtered for profiles, is exactly one? and that one is the profile we use.

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/ScopeAwareProfileResolver.java
enricovianello: Absolutely yes

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/ScopeAwareProfileResolver.java
enricovianello: mm probably `if (clientMatches.size() > 1)` could be enough. If client scopes list is empty you cannot ask other scopes (so yes the intersection is true but in reality the requested ones should be a subset of the others).

enricovianello: ok I got it.
Yes, we can use `clientMatches.size() != 1` 👍

enricovianello: In reality, if no JWT profile matches client scopes (empty case) we can avoid checking the requested scopes (we cannot have more scopes than the allowed ones = no JWT profile scope activators) and return the default profile directly

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/ScopeAwareProfileResolver.java
```diff
@@ -32,22 +33,41 @@ public ScopeAwareProfileResolver(JWTProfile defaultProfile, Map<String, JWTProfi
...
+    Set<JWTProfile> clientMatches = matches(clientScopes);
+    if (clientMatches.isEmpty()) {
       return defaultProfile;
     }
+    if (clientMatches.size() == 1) {
+      return clientMatches.iterator().next();
+    }
+    // clientMatches.size() > 1
+    Set<JWTProfile> requestedMatches = matches(requestedScopes);
+    if (requestedMatches.isEmpty() || requestedMatches.size() > 1) {
```
giacomini, line 63: here you can have a situation where both `clientMatches` and `requestedMatches` contain multiple profiles. wouldn't it make more sense to intersect the two sets and if the intersection contains only one profile to return that one instead of the default one? for example if `clientMatches` is `aarc,wlcg` and `requestedMatches` is `wlcg,kc`, the chosen profile could be `wlcg`

iam-login-service/src/test/java/it/infn/mw/iam/test/oauth/profile/ScopeAwareProfileResolverTests.java
```diff
@@ -118,5 +118,57 @@ public void multipleProfilesLeadToDefaultProfile() {
...
+
+    profile = profileResolver.resolveProfile(clientScopes, Set.of("openid", "aarc"));
+    assertThat(profile, is(aarcProfile));
+
+    profile = profileResolver.resolveProfile(clientScopes, Set.of("openid"));
+    assertThat(profile, is(iamProfile));
+
+    profile = profileResolver.resolveProfile(clientScopes, Set.of("openid", "wlcg", "aarc"));
+    assertThat(profile, is(iamProfile));
+
```
giacomini, line 142: based on the comment above, I would suggest the following test, if you agree with that outcome and it is not a backwards compatibility
```suggestion
    profile = profileResolver.resolveProfile(setOf("aarc", "wlcg"), Set.of("openid", "aarc", "kc"));
    assertThat(profile, is(aarcProfile));
```
(assuming "kc" is an actual profile)

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/ScopeAwareProfileResolver.java
enricovianello: There are filters before that avoid this case. But just to be sure there are not failures I can add an IllegalStateException raised in case of `requestedScopes` is not a subset of `clientScopes`.
The case of both the lists contain multiple profiles:
```
clientScopes = Set.of("openid", "wlcg", "aarc", "kc")
requestedScopes = Set.of("wlcg", "kc")
```
currently is managed returning the default JWT profile (configurable).
If we plan a "merged" outcome we need to discuss and design it.

iam-login-service/src/test/java/it/infn/mw/iam/test/oauth/profile/ScopeAwareProfileResolverTests.java
enricovianello: I'll design this with an expected Exception (kc is not part of client's scopes)
but I'll also add a test with 

```
profile = profileResolver.resolveProfile(setOf("aarc", "wlcg", "kc"), Set.of("aarc", "kc"));
assertThat(profile, is(defaultProfile));
```

But this can be part of a discussion as written above.

enricovianello: Done

iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/profile/ScopeAwareProfileResolver.java
enricovianello: In case the profile resolved is not allowed by client an IllegalArgumentException is thrown.

PR #1117, opened 2025-12-18
Fix for populating fields via idp

### LINKED ISSUE #1116: Issue with auto filling registration fields from IdP
[Pull request 984](https://github.com/indigo-iam/iam/pull/984) introduced the feature of linking a certificate as part of the configuration registration process. 

With this there is a misnaming within the code. 

From `iam-login-service/src/main/resources/application-registration.yml` the field is configured as `certificate`. The same name is persistent, within the back-end code, but when it comes to the front-end it was named `registerCertificate`. This misalignment means that the method `populateFieldsWithAdminPreference` fails as it looks for the field `certificate`, when it's instead called `registerCertificate`. 

As `populateFieldsWithAdminPreference` fails within `activate`, then it never enters the method `populateRequest`. 
This can be confirmed, by entering a new value within one of the fields and clicking `reset-form`. This calls the `populateRequest` and populates the fields according to the information received from the IdP. 

The issue is similar to [Issue 1104](https://github.com/indigo-iam/iam/issues/1104).

S203932:
I changed the name of the field within the front-end to from `registerCertificate` to `certificate`. Thus, being identically within the one defined in the back-end.

### DIFF
--- a/iam-login-service/src/main/java/it/infn/mw/iam/authn/x509/IamX509PreauthenticationProcessingFilter.java
+++ b/iam-login-service/src/main/java/it/infn/mw/iam/authn/x509/IamX509PreauthenticationProcessingFilter.java
@@ -15,9 +15,14 @@
  */
 package it.infn.mw.iam.authn.x509;
 
+import static it.infn.mw.iam.config.IamProperties.ExternalAuthAttributeSectionBehaviour.HIDDEN;
+
 import java.io.IOException;
+import java.security.cert.X509Certificate;
+import java.util.Arrays;
 import java.util.Calendar;
 import java.util.Date;
+import java.util.Objects;
 import java.util.Optional;
 
 import javax.servlet.ServletException;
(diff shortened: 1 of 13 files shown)


S203932:
What's being fixed with this pull request:

1. populating the registrations fields with the Idp information
2. OIDC and SAML can now also use this feature (the X509 filter wasn't present when doing Idp registration and it didn't populate the necessary session attributes)
3. Error's from the certificate expiration pop-up window are now being caught
4. The `certificate` field can be omitted from the `application-registration.yml` (along with all the other fields)

One thing to take note off is that the `certificate` field is part of the registration fields and therefore inherits all the properties of `RegistrationFieldProperties`. This means that one can configure `readOnly` and `fieldBehaviour` for certificate, though there is no benefit from this. Also, it should be noted that one should not set `readOnly = true` for `certificate` since this means that it is the Idp to provide the information about whether the certificate should be linked or not. If one does set this value to true, then it would most likely result in the user not being able to register as the Idp is not providing this required information for the registration. 

Please advise, if any changes are needed.

enricovianello (approved):

PR #1124, opened 2026-01-09
Fix registration request from external IdP

federicaagostini:
Here we fix the registration request when the user was authenticated with an external provider.

In the next example I have logged-into IAM with iam-dev, and the "Given Name" field is read-only

image

Previously, the fields were not populated at all.


federicaagostini:
Not sure if the fix was intended to be included in PR https://github.com/indigo-iam/iam/pull/1117. In case, feel free to close this PR.

federicaagostini:
Closing as the fix looks like included in the linked PR.

PR #1129, opened 2026-01-16
Remove openid-connect-server dependency from login-service module

enricovianello:


enricovianello:
A new approach will be used. These things will be introduced PR by PR. Next ones are #1238 and #1235

PR #1136, opened 2026-01-20
Fix duplicate key error on case-insensitive info_key

### LINKED ISSUE #1140: Duplicate key error when IdP sends duplicate SAML attributes
This is caused by the way SAML attributes are mapped in the INDIGO IAM codebase and by the case-insensitive `info_key` column in the `saved_user_auth_info` db table.

rmiccoli:


iam-persistence/src/main/resources/db/migration/mysql/V115__alter_saved_user_auth_info_key.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE saved_user_auth_info MODIFY info_key varchar(256) CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_bin;
```
rmiccoli: **Note**: `utf8mb4_0900_bin` is MySQL 8.0–only and will fail on MariaDB.
MariaDB provides `utf8mb4_nopad_bin` instead.
The only collation supported by both MySQL 8 and MariaDB is `utf8mb4_bin`, which uses PAD SPACE semantics (trailing spaces are ignored in comparisons).

enricovianello (changes requested):
Switch to `utf8mb4_bin`

enricovianello (approved):

PR #1142, opened 2026-01-23
Fix upper case conversion for TokenTypeHint

### LINKED ISSUE #1139: Incorrect introspection endpoint case sensitivity for TokenTypeHint cause failure at token introspection.
### Description

The IAM currently fails to process OAuth2 introspection requests if the `token_type_hint` parameter is provided in lowercase. While the current implementation expects uppercase values (e.g., `ACCESS_TOKEN`), [[RFC 7009 Section 4.1.2](https://www.rfc-editor.org/rfc/rfc7009#section-4.1.2)](https://www.rfc-editor.org/rfc/rfc7009#section-4.1.2) specifies that the hint values are lower case.

### Affected Code

version > 1.13.x (tested on v1.13.3)

https://github.com/indigo-iam/iam/blob/2419f29517004c6df34d3828b24d72fd5fed3788/iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/introspection/model/TokenTypeHint.java#L18-L21

### RFC Reference

Per **RFC 7009, Section 4.1.2**:

> "The client MAY pass this information as a hint... The following values are defined for this parameter:
> * `access_token`: The token presented for revocation is an access token.
> * `refresh_token`: The token presented for revocation is a refresh token.
> 
> 
> **The values are case-insensitive/ lower cased.**"

### Steps to Reproduce

1. Send a POST request to the revocation or introspection endpoint.
2. Include the parameter `token_type_hint=access_token` (lowercase).
3. The server returns a `400 Bad Request` because it expects `ACCESS_TOKEN`.

### Expected Behavior

The service should accept both `access_token`/`refresh_token` and `ACCESS_TOKEN`/`REFRESH_TOKEN` or just the lower case version as valid hints .

### Suggested Fix

Either make it case insensitive or per RFC document which is lower case.

```java
public enum TokenTypeHint {
  access_token, refresh_token;
}

```

enricovianello:
Add case sensitive test

### DIFF
--- a/iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/introspection/model/TokenTypeHintConverter.java
+++ b/iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/introspection/model/TokenTypeHintConverter.java
@@ -1,3 +1,18 @@
+/**
+ * Copyright (c) Istituto Nazionale di Fisica Nucleare (INFN). 2016-2021
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


DonaldChung-HK:
Hi @enricovianello,

Thanks for the prompt fix.

Thanks,

Donald

PR #1146, opened 2026-01-29
Stop redirection to AUP page for pre authenticated user

### LINKED ISSUE #1122: Unable to sign AUP if SSO login user enabled MFA due to faulty redirect rule
### Problem Description
There appears to be an issue with the Spring Security rule handling the AUP (Acceptable Use Policy) redirect page. 

**Scenario:**
- If an SSO user has an expired AUP, their account becomes disabled.
- The redirect flow is as follows:

AUP → Redirect to Login → SSO Authentication Successful → TOTP →  Redirect to AUP without completing the flow
- Instead of showing the TOTP (MFA) field, the user is sent directly to the AUP page.
- This results in the user getting stuck with an **Access Denied** error because MFA has not been completed.
- Consequently, the user cannot re-sign the AUP to re-enable their account.

### Root Cause
The endpoint for TOTP verification (`/iam/verify`) is behind the rule that forces users to sign the AUP.  
I believe this is the result of differences between how the `PRE-AUTH` role is handled
As a result:
- Users are blocked and prematurely redirected to the AUP page.
- MFA cannot be completed before AUP signing, causing the access flow to break.

### Impact
- Users with expired AUP cannot complete MFA.
- They are unable to re-sign the AUP and regain access.

---

**Suggested Fix:**  
Review and adjust the Spring Security configuration so that MFA verification (`/iam/verify`) is accessible before enforcing AUP signing redirect rule for pre-authenticated user.

garaimanoj:
- A new condition was added to `EnforceAupFilter` so that, when MFA is active for an SSO user, the filter redirects them to the MFA verification page as the first step, instead of sending them to the AUP signing page.
- Test cases for OIDC and SAML flow
- LocalAuthentication is working fine as `accountUtils.isAuthenticated()` is `false` if TOTP not verified

### DIFF
--- a/iam-login-service/src/main/java/it/infn/mw/iam/api/account/AccountUtils.java
+++ b/iam-login-service/src/main/java/it/infn/mw/iam/api/account/AccountUtils.java
@@ -66,6 +66,11 @@ public class AccountUtils {
     return account.getAuthorities().contains(ROLE_ADMIN);
   }
 
+  public boolean isPreAuthenticated() {
+    Authentication auth = SecurityContextHolder.getContext().getAuthentication();
+    return isPreAuthenticated(auth);
+  }
+
   public boolean isPreAuthenticated(Authentication auth) {
     if (auth == null || auth.getAuthorities().isEmpty()) {
       return false;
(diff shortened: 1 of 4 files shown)


rmiccoli (approved):
LGTM 👍

PR #1164, opened 2026-02-15
Add Dockerfile for voms-aa

federicaagostini:
We do not use mvn buildpacks as it was broken and also iam-login-service and iam-test-client were built with their own Dockerfile. Here I add a similar Dockerfile for iam-voms-aa, backward compatible with current deployments. Just the image name is changed from `indigoiam/voms-aa-bp` to `indigoiam/iam-voms-aa`.

Tested with a local build and docker-compose, just changing the image name it worked.


iam-voms-aa/docker/Dockerfile.prod
```diff
@@ -0,0 +1,16 @@
+FROM eclipse-temurin:17 as builder
+RUN mkdir /workspace
```
giacomini: Not needed, see https://docs.docker.com/reference/dockerfile/#workdir

iam-voms-aa/docker/Dockerfile.prod
```diff
@@ -0,0 +1,16 @@
...
+WORKDIR /workspace
+COPY iam-voms-aa.jar /workspace/
+RUN java -Djarmode=layertools -jar iam-voms-aa.jar extract
+
+FROM eclipse-temurin:17
+RUN mkdir /workspace
```
giacomini: Not needed, see https://docs.docker.com/reference/dockerfile/#workdir

iam-voms-aa/docker/Dockerfile.prod
```diff
@@ -0,0 +1,16 @@
...
+WORKDIR /workspace
+
+COPY --from=builder workspace/dependencies/ ./
+COPY --from=builder workspace/spring-boot-loader/ ./
+COPY --from=builder workspace/snapshot-dependencies/ ./
+COPY --from=builder workspace/application/ ./
```
giacomini: To minimize the number of layers, wouldn't it be better to copy all directories in one COPY instruction?
```suggestion
COPY --from=builder /workspace/dependencies/ /workspace/spring-boot-loader/ /workspace/snapshot-dependencies/ /workspace/application/ /workspace/
```
Here I have also used absolute paths both for the sources and the destination. I think it's clearer and safer.

enricovianello (approved):

PR #1111, opened 2025-12-08
Add filter to enforce MFA

garaimanoj:
After log in it will ask to activate MFA -> Once MFA is activated it will redirect to logout page
To access the application user need to login again using MFA.
image

User role is set to `PRE_AUTHENTICATED` when redirected to MFA activation page

PR #1166, opened 2026-02-16
MFA enhancements

enricovianello:
* Allow encryption key rotation (source #1048)
* Stop redirection to AUP page for pre authenticated user (source #1146)
* Add filter to enforce MFA (source: #1111)


enricovianello:
The involved PR have been/will be merged singularly

PR #1184, opened 2026-03-09
Add a mechanism for group membership reminder

### LINKED ISSUE #922: Send Group managers reminders of un-actioned Group requests
If a Group membership request goes un-actioned for a configurable number of days, a reminder email should be automatically sent to the Group Managers.

This would save manual effort from being spent chasing Group membership requests.

Sae126V:
Resolves  #922


iam-login-service/src/main/java/it/infn/mw/iam/core/web/group/GroupRequestReminderTask.java
```diff
@@ -0,0 +1,155 @@
...
+
+    if (thresholdDays < 1 || repeatIntervalDays < 1) {
+      LOG.warn(
+          "Group request reminder is enabled but misconfigured (threshold-days={}, repeat-interval-days={}); "
+              + "both must be >= 1. Skipping.",
+          thresholdDays, repeatIntervalDays);
+      return;
+    }
+
+    Instant now = Instant.now();
```
enricovianello: We need to use the autowired Clock and use clock.instant()

This should help you also on testing it.

iam-login-service/src/main/resources/application.yml
```diff
@@ -226,6 +226,13 @@ task:
   approvalCleanupPeriodMsec: ${IAM_APPROVAL_CLEANUP_PERIOD_MSEC:300000}
   deviceCodeCleanupPeriodMsec: ${IAM_DEVICE_CODE_CLEANUP_PERIOD_MSEC:300000}
   wellKnownCacheCleanupPeriodSecs: ${IAM_WELL_KNOWN_CACHE_CLEANUP_PERIOD_SECS:300}
+  groupRequestReminderPeriodSecs: ${IAM_GROUP_REQUEST_REMINDER_PERIOD_SECS:86400}
+
+group-request-reminder:
+  enabled: ${IAM_GROUP_REQUEST_REMINDER_ENABLED:false}
+  threshold-days: ${IAM_GROUP_REQUEST_REMINDER_THRESHOLD_DAYS:5}
+  repeat-interval-days: ${IAM_GROUP_REQUEST_REMINDER_REPEAT_INTERVAL_DAYS:2}
+  notify-admins: ${IAM_GROUP_REQUEST_REMINDER_NOTIFY_ADMINS:false}
```
enricovianello, line 243: should this property true in order to keep the same behavior without this fix? 
If yes, I'd put true as default value

enricovianello (changes requested):
Minor changes.

iam-login-service/src/main/java/it/infn/mw/iam/core/web/group/GroupRequestReminderTask.java
Sae126V: Cool. Done

iam-login-service/src/main/resources/application.yml
Sae126V: Hmm! This is slightly new functionality, where we remind group managers to action membership requests after `threshold-days` has passed. By default, this ideally notifies only group managers - if there are no group managers, it falls back to notifying admins. When group managers exist, notify-admins is an option to loop in the admins as well when we send the reminder email about pending membership requests.

PR #1185, opened 2026-03-09
Make new dashboard client credentials configurable on bootstrap

SteDev2:
This PR introduces the ability to configure/create a client (id + secret) credential used by the [new IAM dashboard]().
Administrators can configure this feature by adding:

```
iam:
  dashboard:
    enabled: true
    client-id: my-client-id
    client-secret: my-v@liD-secret
```

Or they can use the environment variables:

```
IAM_DASHBOARD_ENABLED = true
IAM_DASHBOARD_CLIENT_ID = my-client-id
IAM_DASHBOARD_CLIENT_SECRET = my-v@liD-secret
```
 
At each IAM restart, if `iam.dashboard.enabled` is true, if it's not present, a client with those id and secret is created. In case the client exists, things like grant types, authentication method, scopes, redirect URIs, etc. are checked to be compliant for a "dashboard client" setup and if not, they're updated properly. If all the values are fine, the record is ignored.

This PR adds also an important limit to the length of user passwords and client secrets: 72 characters. This is due to a safer usage of b-crypt encryption.


iam-login-service/src/main/java/it/infn/mw/iam/config/IamProperties.java
```diff
@@ -599,6 +602,39 @@ public void setTrackLastUsed(boolean trackLastUsed) {
     }
   }
 
+  @Validated
+  @ConfigurationProperties(prefix = "app.dashboard")
```
enricovianello: Maybe useless

iam-login-service/src/main/java/it/infn/mw/iam/core/util/StartupRunner.java
```diff
@@ -0,0 +1,58 @@
...
+  @Override
+  public void run(ApplicationArguments args) {
+    if (!dashboardProperties.isEnabled()) {
+      LOG.info(
+          "Dashboard client is disabled, skipping checks for the dashboard client properties and the presence of the record for the dashboard client");
+      return;
+    }
+
+    boolean recordExists = dashboardConfigService.initDashboardClient(dashboardProperties, iamBaseUrl);
+    if (!recordExists) {
```
enricovianello: success instead of recordExists?

iam-login-service/src/main/java/it/infn/mw/iam/dashboard/DashboardConfigService.java
```diff
@@ -0,0 +1,134 @@
...
+  private final IamClientRepository clientRepository;
+  private final DefaultClientManagementService clientService;
+
+  public DashboardConfigService(IamClientRepository clientRepository,
+      DefaultClientManagementService clientService) {
+    this.clientService = clientService;
+    this.clientRepository = clientRepository;
+  }
+
+  public boolean initDashboardClient(DashboardProperties dashboardProperties, String iamUrl) {
```
enricovianello: I think "init" as name is sufficient

iam-login-service/src/main/java/it/infn/mw/iam/dashboard/DashboardConfigService.java
```diff
@@ -0,0 +1,134 @@
...
+  private boolean hasAllRequiredScopes(ClientDetailsEntity client) {
+    return client.getScope().containsAll(dashboardScopes);
+  }
+
+  private boolean hasValidClientSecret(ClientDetailsEntity client, String clientSecret) {
+    return client.getClientSecret().equals(clientSecret);
+  }
+
+  private boolean hasValidRedirectUris(ClientDetailsEntity client, String url) {
+    return client.getRedirectUris().containsAll(Set.of(url));
```
enricovianello: You could use `client.getRedirectUris().retainAll(Set.of(url)) == false`
means valid because a true response means the retainAll method has changed the list of redirect URIs

This approach changes the way we're validating that list in the sense that we're NOT accepting multiple redirectURIs

iam-login-service/src/main/java/it/infn/mw/iam/dashboard/DashboardConfigService.java
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  private boolean supportsAuthorizationCodeGrant(ClientDetailsEntity client) {
+    return client.getGrantTypes().containsAll(
```
enricovianello: we can apply the same logic above

iam-login-service/src/main/java/it/infn/mw/iam/dashboard/DashboardConfigService.java
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  public boolean initDashboardClient(DashboardProperties dashboardProperties, String iamUrl) {
+    String clientId = dashboardProperties.getClientId();
+    String clientSecret = dashboardProperties.getClientSecret();
+    String url = iamUrl + DASHBOARD_CALLBACK;
+
+    Optional<ClientDetailsEntity> test = clientRepository.findByClientId(clientId);
+
+    if (test.isPresent()) {
```
enricovianello: I'd remove the else and add something like:

```
if (...) {
  return;
}

// else
```

iam-login-service/src/main/java/it/infn/mw/iam/dashboard/DashboardConfigService.java
```diff
@@ -0,0 +1,134 @@
...
+      ClientDetailsEntity client = test.get();
+
+      boolean isConfigured = checkRecordConfiguration(test.get(), clientSecret, url);
```
enricovianello: boolean isValid = validateConfiguration() ?

enricovianello (changes requested):
Minor fixes

iam-login-service/src/main/java/it/infn/mw/iam/core/util/StartupRunner.java
SteDev2: i change with isValid

iam-login-service/src/main/java/it/infn/mw/iam/dashboard/DashboardConfigService.java
SteDev2: It's a good idea for scopes but not for RedirectUrl and GrantType which must be exactly identical, I preferred to use equals

iam-login-service/src/main/java/it/infn/mw/iam/core/util/StartupRunner.java
```diff
@@ -0,0 +1,52 @@
...
+  @Override
+  public void run(ApplicationArguments args) {
+    if (!dashboardConfigService.isEnabled()) {
+      LOG.info(
+          "Dashboard client is disabled, skipping checks for the dashboard client properties and the presence of the record for the dashboard client");
+      return;
+    }
+
+    boolean isValid = dashboardConfigService.init();
+    if (!isValid) {
```
enricovianello: checking the code of the service, I see `false` is returned after an Exception is thrown on db entry creation. I think we can raise another error here saying something like "Error during dashboard client initialization. Startup failed. Please check the logged error."

iam-login-service/src/main/java/it/infn/mw/iam/core/util/StartupRunner.java
```diff
@@ -0,0 +1,52 @@
...
+
+  public StartupRunner(DashboardConfigService dashboardConfigService) {
+    this.dashboardConfigService = dashboardConfigService;
+  }
+
+  @Override
+  public void run(ApplicationArguments args) {
+    if (!dashboardConfigService.isEnabled()) {
+      LOG.info(
+          "Dashboard client is disabled, skipping checks for the dashboard client properties and the presence of the record for the dashboard client");
```
enricovianello: I'd simplify this message into "Skipping dashboard client initialization." The perfect message should include something like "Read more info about here: " + URL of an updated IAM documentation.

iam-login-service/src/main/java/it/infn/mw/iam/dashboard/DashboardConfigService.java
```diff
@@ -0,0 +1,155 @@
...
+    this.iamProperties = iamProperties;
+  }
+
+  public boolean isEnabled() {
+    return iamProperties.getDashboard().isEnabled();
+  }
+
+  public boolean init() {
+    DashboardProperties dashboardProperties = iamProperties.getDashboard();
+    String iamUrl = iamProperties.getBaseUrl();
```
enricovianello, line 77: This thing is really relevant to be reported on the IAM related documentation. The dashboard MUST be on the same IAM hostname in order to be configured by this logic. Is this what we really want @giacomini ? Or could we allow having a configurable list of redirect URIs in order to allow admins to deploy a new dashboard on a different host?

iam-login-service/src/main/java/it/infn/mw/iam/dashboard/DashboardConfigService.java
```diff
@@ -0,0 +1,155 @@
...
+    String clientId = dashboardProperties.getClientId();
+    String clientSecret = dashboardProperties.getClientSecret();
+    String url = iamUrl + DASHBOARD_CALLBACK;
+    Optional<ClientDetailsEntity> dashboardRecord = clientRepository.findByClientId(clientId);
+
+    if (!dashboardRecord.isPresent()) {
+      LOG.info("The client record for dashboard does not exist. Creating record with default configuration...");
```
enricovianello: "Dashboard client does not exist and it will be created."

iam-login-service/src/main/java/it/infn/mw/iam/dashboard/DashboardConfigService.java
```diff
@@ -0,0 +1,155 @@
...
+      } catch (Exception e) {
+        LOG.error("Error saving dashboard client: {}", e.getMessage());
+        return false;
+      }
+    }
+
+    ClientDetailsEntity client = dashboardRecord.get();
+    boolean isValid = checkRecordConfiguration(client, clientSecret, url);
+    if (!isValid) {
+      LOG.warn("The record is not properly configured. Updating Dashboard client.");
```
enricovianello: "Changes on default dashboard client configuration found: restoring expected configuration."

I know it's a database record but I'd like to communicate an higher level concept to the final admins :)

iam-login-service/src/main/java/it/infn/mw/iam/dashboard/DashboardConfigService.java
```diff
@@ -0,0 +1,155 @@
...
+  private boolean hasAllRequiredScopes(ClientDetailsEntity client) {
+    return client.getScope().containsAll(DASHBOARD_SCOPES);
+  }
+
+  private boolean hasValidClientSecret(ClientDetailsEntity client, String clientSecret) {
+    return client.getClientSecret().equals(clientSecret);
+  }
+
+  private boolean hasValidRedirectUris(ClientDetailsEntity client, String url) {
+    return client.getRedirectUris().equals(Set.of(url));
```
enricovianello: I think inverting things could be safer:

`Set.of(url).equals(client.getRedirectUris())`
 
this protects us from null pointers

iam-login-service/src/main/java/it/infn/mw/iam/dashboard/DashboardConfigService.java
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  private boolean supportsAuthorizationCodeGrant(ClientDetailsEntity client) {
+    return client.getGrantTypes().equals(
+        Set.of(AuthorizationGrantType.CODE.getGrantType(), AuthorizationGrantType.REFRESH_TOKEN.getGrantType()));
```
enricovianello: same as above: the static set should be the one

iam-login-service/src/main/java/it/infn/mw/iam/dashboard/DashboardConfigService.java
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  private boolean usesClientSecretBasicAuth(ClientDetailsEntity client) {
+    return client.getTokenEndpointAuthMethod().equals(AuthMethod.SECRET_BASIC);
```
enricovianello: same as above

iam-login-service/src/main/java/it/infn/mw/iam/dashboard/DashboardConfigService.java
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  private boolean usesPKCES256(ClientDetailsEntity client) {
+    return client.getCodeChallengeMethod().getName().equals(PKCEAlgorithm.S256.toString());
```
enricovianello: same as above

iam-login-service/src/main/java/it/infn/mw/iam/dashboard/DashboardConfigService.java
```diff
@@ -0,0 +1,155 @@
...
+
+    clientRepository.save(client);
+  }
+
+  private boolean hasAllRequiredScopes(ClientDetailsEntity client) {
+    return client.getScope().containsAll(DASHBOARD_SCOPES);
+  }
+
+  private boolean hasValidClientSecret(ClientDetailsEntity client, String clientSecret) {
+    return client.getClientSecret().equals(clientSecret);
```
enricovianello: same as below

iam-login-service/src/main/java/it/infn/mw/iam/dashboard/DashboardConfigService.java
```diff
@@ -0,0 +1,155 @@
...
+    client.setCodeChallengeMethod(PKCEAlgorithm.S256);
+    client.setClientSecret(secret);
+    client.setRedirectUris(Set.of(url));
+    client.setTokenEndpointAuthMethod(AuthMethod.SECRET_BASIC);
+
+    clientRepository.save(client);
+  }
+
+  private boolean hasAllRequiredScopes(ClientDetailsEntity client) {
+    return client.getScope().containsAll(DASHBOARD_SCOPES);
```
enricovianello: I'd add a safer:

`client.getScope() != null &&` before

to protect from null pointer exceptions

iam-login-service/src/main/java/it/infn/mw/iam/dashboard/DashboardConfigService.java
```diff
@@ -0,0 +1,155 @@
...
+  private void updateRecordDashboard(ClientDetailsEntity client, String secret, String url) {
+    client.setScope(DASHBOARD_SCOPES);
+    client.setGrantTypes(
+        Set.of(AuthorizationGrantType.CODE.getGrantType(), AuthorizationGrantType.REFRESH_TOKEN.getGrantType()));
+    client.setCodeChallengeMethod(PKCEAlgorithm.S256);
+    client.setClientSecret(secret);
+    client.setRedirectUris(Set.of(url));
+    client.setTokenEndpointAuthMethod(AuthMethod.SECRET_BASIC);
+
+    clientRepository.save(client);
```
enricovianello, line 133: We're not using client service so we need to add a ClientUpdatedEvent here

enricovianello: Or simply use clientService.updateClient() which also evicts cached values

enricovianello (changes requested):
Probably some safer comparisons could be done and a missing ClientUpdatedEvent to be fixed. Also minor things about log messages.
But from the point of view of the logic it's ok.
Theres's also a question for @giacomini about supporting or not a configurable list of redirect URIs in order to allow different hostnames (different from the issuer).

iam-login-service/src/main/java/it/infn/mw/iam/dashboard/DashboardConfigService.java
jacogasp: > The dashboard MUST be on the same IAM hostname in order to be configured by this logic.

Right now, the dashboard MUST be on the same hostname only because of logout issue (i.e., cookies deletion). This requirement would be superseded by [this PR](https://github.com/indigo-iam/iam/pull/1180).

In such case the above PR is approved and merged, this will not be mandatory anymore.

In my personal opinion, I would NOT let user to choose anything. Variables like

- hostname
- dashboard base path (`/ui`)
- dashboard redirect uri

should be hardcoded (possibly also the client name/id, for example `iam-default-dashboard`).

The secret is the only thing that the user should manually insert.

> list of redirect URIs in order to allow admins to deploy a new dashboard on a different host

Don't know if the having the same client for different redirect_uri (i.e., one client for many dashboards) is a good idea. Messing up with a single client could potentially break all dashboards.

To recap, still in my personal opinion: the default client is IMMUTABLE with redirect uri `<IAM_HOSTNAME>/ui/api/oauth2/indigo-iam`.
For every other dashboard, the admin can manually create a client as they would do for whatever web application.

giacomini: I agree with @jacogasp: for the time being let's minimize the configuration, which means just client id and client secret

enricovianello: Thanks for your clarifications. So, we can proceed as it is now implemented by this PR.

iam-login-service/src/main/java/it/infn/mw/iam/dashboard/DashboardConfigService.java
SteDev2: clientService.updateClient() requier a clientDTO object, I prefer add ClientUpdatedEvent on this part of code.

enricovianello (approved):

PR #1187, opened 2026-03-16
Bypass auth check in case of no auth client

### LINKED ISSUE #1175: Verification of behaviour of Token endpoint authentication method
Hi,

We have user report that changing the Token endpoint authentication method from `Client secret with symmetrically signed JWT assertion` seems to behave the same as `Client secret over HTTP POST authentication`. Thus, please verify if the current MitreID behavior have a default to just accepting the Client_ID and Client_Secret if it sees it in the content of the link.

e.g. a jtest token endpoint authentication set to `Client secret with symmetrically signed JWT assertion` but with the valid client_id and client_secret (especially a client created by user) in HTTP header or HTTP body to check if it can validate.

In addition to that, users have reported that changing a client with existing client secret to no authentication needs the existing client secret, while a brand new client that never have any client secret set can be authenticated with No auth as it is still `""`

garaimanoj:


iam-login-service/src/main/java/it/infn/mw/iam/authn/PublicClientAwareCcFilter.java
```diff
@@ -0,0 +1,65 @@
...
+import org.springframework.security.authentication.UsernamePasswordAuthenticationToken;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.oauth2.provider.client.ClientCredentialsTokenEndpointFilter;
+
+import it.infn.mw.iam.api.client.service.ClientService;
+
+public class PublicClientAwareCcFilter extends ClientCredentialsTokenEndpointFilter {
+
+    private final ClientService clientService;
```
rmiccoli, line 38: I'd use `IamClientRepository` class and its `findByClientId` method.

rmiccoli (changes requested):

iam-login-service/src/main/java/it/infn/mw/iam/authn/PublicClientAwareCcFilter.java
garaimanoj: Using the service layer in a filter keeps business logic centralized and ensures consistency across the application. It avoids tightly coupling the filter to the database layer and allows validation, rules, and future changes to be handled in one place, improving maintainability and testability.

rmiccoli: I agree. It was simply a matter of not using the MITREid service, but I'll let @enricovianello comment on that.

rmiccoli (approved):

PR #1191, opened 2026-03-23
Make SAML authn-context class-refs list configurable

### LINKED ISSUE #1130: Make AuthnContext in SAML Requests Customisable or Optional
### Summary
The `AuthnContextClassRef` values in SAML authentication requests are currently fixed. While the original defaults worked well with traditional Shibboleth-based IdPs, many modern IdPs now deviate from strict Shibboleth semantics. As a result, some users are unable to authenticate, particularly when using proxied identity providers such as Microsoft Entra ID.

Linking #1012 

### Background / Context
Historically, the predefined list of `AuthnContextClassRef` values aligned well with Shibboleth IdPs and their interpretations. However, in recent years:

- Some IdP don't use Shibboleth, some proxied in front of Shibboleth.
- There has been a significant move towards proxied IdPs, for example, Microsoft Entra ID acting as a proxy to leverage built-in MFA/2FA
- These proxied setups often reject authentication when a required or unsupported AuthnContext is included

This results in valid users from those organisation being unable to log in, even though authentication and assurance requirements are otherwise met.

### Problem
- The current SAML implementation always sends a fixed list of `AuthnContextClassRef` values. See below
https://github.com/indigo-iam/iam/blob/5fd64750f90e90453753e33e0a2bc8b86e289578/iam-login-service/src/main/java/it/infn/mw/iam/authn/saml/profile/IamSSOProfile.java#L87-L90
- Some IdPs has a custom set that no longer supports the usual `PasswordProtectedTransport`
- This particularly affects organisations using Entra ID proxied IdP or where they don't use Shibboleth.

### Impact
- Users are unable to SSO authenticate into INDIGO IAM and requires a local account
- Increased support and operational overhead

### Suggested Fix
One or both of the following:

1. Allow `AuthnContextClassRef` to be fully configurable
   - Admin-defined list
   - Per-deployment or per-SAML item configuration (Optional)

2. Allow sending no `AuthnContextClassRef` at all (Preferred if we are moving to OIDC and do not want to spend time debugging this)
   - Do not include `AuthnContext` in the SAML request
   - Let the IdP determine the appropriate authentication assurance

Sae126V:
Resolves #1130

### DIFF
--- a/iam-login-service/src/main/java/it/infn/mw/iam/authn/saml/profile/IamSSOProfile.java
+++ b/iam-login-service/src/main/java/it/infn/mw/iam/authn/saml/profile/IamSSOProfile.java
@@ -15,10 +15,6 @@
  */
 package it.infn.mw.iam.authn.saml.profile;
 
-import java.util.List;
-import java.util.Set;
-import java.util.stream.Collectors;
-
 import javax.xml.namespace.QName;
 
 import org.opensaml.common.SAMLException;
@@ -84,7 +80,7 @@ public class IamSSOProfile extends WebSSOProfileImpl {
 
   private void configureAuthnContext(AuthnRequest request) {
 
-    if (!authnContextProperties.isEnabled()) {
+    if (!authnContextProperties.isDefined()) {
       // Actively clear any AuthnContext the parent class may have already set.
       // This ensures NO RequestedAuthnContext is sent letting the IdP decide.
       request.setRequestedAuthnContext(null);
--- a/iam-login-service/src/main/java/it/infn/mw/iam/authn/saml/profile/IamSSOProfile.java
+++ b/iam-login-service/src/main/java/it/infn/mw/iam/authn/saml/profile/IamSSOProfile.java
@@ -15,8 +15,11 @@
  */
 package it.infn.mw.iam.authn.saml.profile;
 
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
 import javax.xml.namespace.QName;
-import static java.util.Objects.isNull;
 
 import org.opensaml.common.SAMLException;
 import org.opensaml.common.SAMLObject;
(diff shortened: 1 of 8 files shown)


iam-login-service/src/main/resources/application-saml.yml
```diff
@@ -42,4 +42,13 @@ saml:
       metadata-url: ${IAM_SAML_IDP_METADATA:classpath:/saml/idp-metadata.xml}
       require-valid-signature: ${IAM_SAML_METADATA_REQUIRE_VALID_SIGNATURE:false}
       require-sirtfi: ${IAM_SAML_METADATA_REQUIRE_SIRTFI:false}
-      require-rs: ${IAM_SAML_METADATA_REQUIRE_RS:false}
\ No newline at end of file
+      require-rs: ${IAM_SAML_METADATA_REQUIRE_RS:false}
+
+  authn-context:
+    defined: ${IAM_SAML_AUTHN_CONTEXT_DEFINED:true}
```
enricovianello: I'd try to use only class-refs list:
- default: the current list of class-refs
- empty list is allowed

enricovianello (changes requested):

iam-login-service/src/main/java/it/infn/mw/iam/config/saml/IamSamlProperties.java
```diff
@@ -195,6 +196,24 @@ public String mode() {
...
   }
 
+  public static class AuthnContextProperties {
+
+    private List<String> classRefs = Arrays.asList(
+        "https://refeds.org/profile/mfa",
+        "https://refeds.org/profile/sfa",
+        "urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport",
+        "urn:oasis:names:tc:SAML:2.0:ac:classes:unspecified"
+    );
```
DonaldChung-HK: I think I agree with @enricovianello that I prefer not have this list hanging in the source code. It might cause confusion in future developers. 

While the concerns for user accidentally leaving it empty is valid, I think that including it in the default bundled application-saml.yml file should be sufficient.

DonaldChung-HK (changes requested):

iam-login-service/src/main/java/it/infn/mw/iam/config/saml/IamSamlProperties.java
Sae126V: Sure @DonaldChung-HK , @enricovianello. I have made the changes now. All we need to make sure is we have a communicate this loudly that admins need to configure or update the env variables like we spoke `iam-login-service/src/main/resources/application-saml.yml`

enricovianello: I think that it's not possible to configure SAML "things" without using `saml` profile because I saw a lot of dependent beans defined inside `SamlConfig`. So, relying only on the `application-saml.yml` and removing the static list from codebase won't introduce any backward incompatibility 👍

iam-login-service/src/main/java/it/infn/mw/iam/authn/saml/profile/IamSSOProfile.java
```diff
@@ -75,34 +79,28 @@ private <T extends SAMLObject> SAMLObjectBuilder<T> getBuilder(QName elementName
...
 
-  private void addRefedsAuthnContexts(AuthnRequest request) {
-    RequestedAuthnContext requestedAuthnContext = request.getRequestedAuthnContext();
-    SAMLObjectBuilder<RequestedAuthnContext> builder =
-        getBuilder(RequestedAuthnContext.DEFAULT_ELEMENT_NAME);
-    if (requestedAuthnContext == null) {
-      requestedAuthnContext = builder.buildObject();
-      request.setRequestedAuthnContext(requestedAuthnContext);
-    }
+  private void configureAuthnContext(AuthnRequest request) {
```
enricovianello, line 85: The new logic is not considering the existing (requested) classRefs. The old logic was adding the elements of the static list of required class-refs in case they weren't requested. But that logic was preserving the requested ones. And the new logic is also not reproducing the same creation of a default AuthnContext when it's null (lines 79-85).

I'd prefer to consider the configured list of class-refs as the previous "required" one.

Or am i missing something?

Sae126V: How about now, Enrico ?

enricovianello: I think we could simply initialize the old `requiredClassRefs` with the list (empty or not) that comes from the configuration. And leave all the other things as they are. I mean, this new implementation, from line 88 to 93, changes the behavior. If I have no configured classRefs, the requested ones are ignored.

Sae126V: Now <saml2p:RequestedAuthnContext/> will be send but without any child. So that IDP receives and decides whether to go further or not. Are you happy with this, Enrico ? If not, Happy to have a quick call over our channel.

enricovianello: Summarizing:
- having no class refs defined in our configuration means no class refs will be required (this introduces the option of not sending anything)
- empty list means preserve original request
- defined list means you ensure the required by conf are added (default and backward comp.)

I think it's ok. Checking the issue, Donald was asking also an option to totally ignore this.

enricovianello (approved):

enricovianello (approved):

PR #1195, opened 2026-04-08
Remove hamcrest

enricovianello:


enricovianello:
Postponed to future releases

PR #1236, opened 2026-05-07
Change default Dashboard client name and description

enricovianello:


iam-login-service/src/main/java/it/infn/mw/iam/dashboard/DashboardConfigService.java
```diff
@@ -43,6 +43,8 @@ public class DashboardConfigService {
 
   private static final Logger LOG = LoggerFactory.getLogger(DashboardConfigService.class);
 
+  public static final String DASHBOARD_CLIENT_NAME = "The INDIGO IAM Dashboard";
+  public static final String DASHBOARD_CLIENT_DESCRIPTION = "Creating trust, one token at a time";
```
giacomini: Nice try. Let's postpone to when we are very confident that people can really really trust us :-)

PR #1240, opened 2026-05-18
Clean up expired registrations and associated accounts

### LINKED ISSUE #1172: Feature Request: Registration Timeout
We could add a timeout in order to avoid users stuck on a not received confirmation email (because of a failure or a wrong email address). After a certain amount of hours (configurable) the request can be automatically canceled

garaimanoj:
Implemented scheduled cleanup of expired registration requests and associated accounts.

- Added configurable cron-based scheduled task to trigger cleanup
- Introduced expiry logic based on configurable expiryDays
- Ensured safe deletion order (registrations -> accounts) to avoid FK violations
- Optimized service logic with early exit when no expired records are found
- Added unit tests for service and scheduling logic
- Externalized configuration using application properties (cron, enabled)

### DIFF
--- a/iam-login-service/src/main/java/it/infn/mw/iam/config/TaskConfig.java
+++ b/iam-login-service/src/main/java/it/infn/mw/iam/config/TaskConfig.java
@@ -160,7 +160,7 @@ public class TaskConfig implements SchedulingConfigurer {
   }
 
   private void scheduledCleanUpExpireRegistration() {
-    LOG.info("Running registration request cleanup with expiry days= {}", expiryDays);
+    LOG.info("Running expired registration request cleanup with expiry days= {}", expiryDays);
     Instant expiryTime = Instant.now().minus(expiryDays, ChronoUnit.DAYS);
     registrationRequestRepository.findExpiredRequests(Date.from(expiryTime))
         .forEach(registrationRequestService::timeoutRequest);
--- a/iam-login-service/src/main/java/it/infn/mw/iam/config/TaskConfig.java
+++ b/iam-login-service/src/main/java/it/infn/mw/iam/config/TaskConfig.java
@@ -15,6 +15,8 @@
  */
 package it.infn.mw.iam.config;
 
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
 
@@ -40,6 +42,7 @@ import it.infn.mw.iam.core.web.aup.AupReminderTask;
 import it.infn.mw.iam.core.web.wellknown.IamWellKnownInfoProvider;
 import it.infn.mw.iam.notification.NotificationDeliveryTask;
 import it.infn.mw.iam.notification.service.NotificationStoreService;
+import it.infn.mw.iam.registration.RegistrationRequestService;
 import net.javacrumbs.shedlock.core.LockProvider;
 import net.javacrumbs.shedlock.provider.jdbctemplate.JdbcTemplateLockProvider;
 import net.javacrumbs.shedlock.spring.annotation.EnableSchedulerLock;
@@ -61,6 +64,7 @@ public class TaskConfig implements SchedulingConfigurer {
   public static final long ONE_DAY_MSEC = 24 * ONE_HOUR_MSEC;
 
   private NotificationStoreService notificationStoreService;
+  private RegistrationRequestService registrationRequestService;
   private NotificationDeliveryTask deliveryTask;
   private LifecycleProperties lifecycleProperties;
   private ExpiredAccountsHandler expiredAccountsHandler;
@@ -74,12 +78,23 @@ public class TaskConfig implements SchedulingConfigurer {
   @Value("${notification.taskDelay}")
   long notificationTaskPeriodMsec;
 
+  @Value("${iam.registration.expiry-days:7}")
+  private long expiryDays;
+  
+  @Value("${iam.cleanup.expired.registration.cron.enable:true}")
+  private boolean cleanupExpiredRegistrationCronScheduleEnable;
+
+  @Value("${iam.cleanup.expired.registration.cron:0 */2 * * * *}")
+  private String cleanupExpiredRegistrationCronSchedule;
+
   public TaskConfig(NotificationStoreService notificationStoreService,
+      RegistrationRequestService registrationRequestService,
       NotificationDeliveryTask deliveryTask, LifecycleProperties lifecycleProperties,
       ExpiredAccountsHandler expiredAccountsHandler, AupReminderTask aupReminderTask,
       ExecutorService taskScheduler, GarbageCollector garbageCollector) {
 
     this.notificationStoreService = notificationStoreService;
+    this.registrationRequestService = registrationRequestService;
     this.deliveryTask = deliveryTask;
     this.lifecycleProperties = lifecycleProperties;
     this.expiredAccountsHandler = expiredAccountsHandler;
@@ -128,6 +143,22 @@ public class TaskConfig implements SchedulingConfigurer {
   public void scheduledAupRemindersTask() {
 
     aupReminderTask.sendAupReminders();
+  }
+
+  public void scheduledCleanUpExpireRegistrationTask(final ScheduledTaskRegistrar taskRegistrar) {
+    if (!cleanupExpiredRegistrationCronScheduleEnable) {
+      LOG.info("Expired accounts task is disabled");
+    } else {
+      LOG.info("Scheduling expired registration task with schedule: {}", cleanupExpiredRegistrationCronSchedule);
+      taskRegistrar.addCronTask(this::scheduledCleanUpExpireRegistration, cleanupExpiredRegistrationCronSchedule);
+    }
+  }
+
+  // @Scheduled(cron = "0 6 16 * * ?")
+  public void scheduledCleanUpExpireRegistration() {
+    LOG.info("Running cleanup with expiryDays={}", expiryDays);
+    Instant expiryTime = Instant.now().minus(expiryDays, ChronoUnit.DAYS);
+    registrationRequestService.cleanupExpiredRequests(expiryTime);
   }
 
   public void schedulePendingNotificationsDelivery(final ScheduledTaskRegistrar taskRegistrar) {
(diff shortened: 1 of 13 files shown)


iam-persistence/src/main/java/it/infn/mw/iam/persistence/repository/IamRegistrationRequestRepository.java
```diff
@@ -39,4 +41,15 @@ Optional<IamRegistrationRequest> findByAccountConfirmationKey(
 
   @Query("select r from IamRegistrationRequest r where r.status = it.infn.mw.iam.core.IamRegistrationRequestStatus.NEW or r.status = it.infn.mw.iam.core.IamRegistrationRequestStatus.CONFIRMED")
   List<IamRegistrationRequest> findPendingRequests();
+
+  @Query("SELECT r.account.id FROM IamRegistrationRequest r " +
+          "WHERE r.status = it.infn.mw.iam.core.IamRegistrationRequestStatus.NEW " +
+          "AND r.creationTime < :expiryTime")
+  List<Long> findAccountIdsForExpiredRegistrations(@Param("expiryTime") Date expiryTime);
```
enricovianello: Alternatively, we could avoid creating the DELETE Query and use another select here:

```
@Query("SELECT r FROM IamRegistrationRequest r " +
          "WHERE r.status = it.infn.mw.iam.core.IamRegistrationRequestStatus.NEW " +
          "AND r.creationTime < :expiryTime")
  List<IamRegistrationRequest> findExpiredRequests(@Param("expiryTime") Date expiryTime);
```

Then the logic inside the periodic task could be:

```
findExpiredRequests(Date.from(...)).stream().forEach(requestService::timeoutRequest);
```

Of course we need to add the proper service method as described above.

I would change the method name properly, e.g:
`findExpiredRequests`

iam-login-service/src/main/java/it/infn/mw/iam/registration/DefaultRegistrationRequestService.java
```diff
@@ -425,4 +426,23 @@ public RegistrationRequestDto approveRequest(String requestUuid) {
 
     return handleApprove(request);
   }
```
enricovianello, line 430: I'd prefer to implement a method like a:

```
@Override
public RegistrationRequestDto timeoutRequest(String requestUuid) {
  IamRegistrationRequest request = findRequestById(requestUuid);
  if (!checkStatusTransition(request.getStatus(), TIMEOUT)) {
    throw new IllegalArgumentException(
      String.format("Bad status transition from [%s] to [%s]", request.getStatus(), TIMEOUT));
  }
  return handleReject(request, Optional.of("Expired timeout"), true);
}
```

The method `handleReject` can be used also by the scheduled Task, if we change the query that loads the expired requests properly (as described in a comment below).

iam-login-service/src/main/java/it/infn/mw/iam/registration/DefaultRegistrationRequestService.java
```diff
@@ -425,4 +426,23 @@ public RegistrationRequestDto approveRequest(String requestUuid) {
...
+    List<Long> accountIds = requestRepository.findAccountIdsForExpiredRegistrations(expiryDate);
+
+    if (accountIds.isEmpty()) {
+      LOG.debug("Cleanup skipped: no expired registrations found (expiryTime= {})", expiryTime);
+      return;
+    }
+
+    int deletedRegistrations = requestRepository.deleteExpiredRegistrations(expiryDate);
+    int deletedAccounts = accountService.deleteAccountsForExpiredRegistrations(accountIds);
+
```
enricovianello: If we use the alternative strategy described above, this method should become something like:

```
Date expiryDate = Date.from(expiryTime);
List<IamRegistrationRequest> expiredRequests = findExpiredRequests(expiryDate);
expiredRequests.stream().forEach(requestService::timeoutRequest);
LOG.info("Cleanup removed {} registrations (expiryTime= {})", expiredRequests.size(), expiryTime);
```

enricovianello (changes requested):
I proposed you a different approach based on the rejectRequest implementation already present. We can assume the request has been rejected due to a timeout reson in the end.

I apologize. It's a non-small change :)

iam-persistence/src/main/java/it/infn/mw/iam/persistence/repository/IamRegistrationRequestRepository.java
garaimanoj: Done

iam-login-service/src/main/java/it/infn/mw/iam/registration/DefaultRegistrationRequestService.java
garaimanoj: Done

iam-login-service/src/main/java/it/infn/mw/iam/registration/DefaultRegistrationRequestService.java
garaimanoj: Done

iam-login-service/src/main/java/it/infn/mw/iam/config/TaskConfig.java
```diff
@@ -74,12 +81,25 @@ public class TaskConfig implements SchedulingConfigurer {
   @Value("${notification.taskDelay}")
   long notificationTaskPeriodMsec;
 
+  @Value("${iam.cleanup.expired.registration.expiry-days:7}")
```
enricovianello: having included them inside the application.yml, here the default values are not necessary

iam-login-service/src/main/java/it/infn/mw/iam/config/TaskConfig.java
```diff
@@ -74,12 +81,25 @@ public class TaskConfig implements SchedulingConfigurer {
...
+  private long expiryDays;
+  
+  @Value("${iam.cleanup.expired.registration.enabled:false}")
+  private boolean cleanupExpiredRegistrationCronScheduleEnabled;
```
enricovianello, line 88: as above

iam-login-service/src/main/java/it/infn/mw/iam/config/TaskConfig.java
```diff
@@ -74,12 +81,25 @@ public class TaskConfig implements SchedulingConfigurer {
...
+
+  @Value("${iam.cleanup.expired.registration.cron:0 0 7 * * *}")
```
enricovianello: as above

iam-login-service/src/main/java/it/infn/mw/iam/core/user/DefaultIamAccountService.java
```diff
@@ -805,4 +805,27 @@ public IamAccount signAup(IamAccount account, IamAup aup) {
     return account;
   }
 
+  @Override
+  public int deleteAccountsForExpiredRegistrations(List<Long> ids) {
```
enricovianello: This should be useless now. It's used only by tests.

iam-login-service/src/main/java/it/infn/mw/iam/core/user/IamAccountService.java
```diff
@@ -255,4 +256,13 @@ IamAccount createAccount(RegistrationRequestDto dto,
...
   IamAccount signAup(IamAccount account, IamAup aup);
+
+  
+  /**
+   * Deletes {@link IamAccount} records associated with the given account IDs.
+   *
+   * @param ids list of account IDs to delete
+   * @return number of accounts successfully deleted
+   */
+  int deleteAccountsForExpiredRegistrations(List<Long> ids);
```
enricovianello: Removable because not used (except for tests)

iam-login-service/src/main/java/it/infn/mw/iam/config/TaskConfig.java
```diff
@@ -130,6 +150,22 @@ public void scheduledAupRemindersTask() {
...
+      LOG.info("Scheduling expired registration request clean up task with schedule: {}", cleanupExpiredRegistrationCronSchedule);
+      taskRegistrar.addCronTask(this::scheduledCleanUpExpireRegistration, cleanupExpiredRegistrationCronSchedule);
+    }
+  }
+
+  private void scheduledCleanUpExpireRegistration() {
+    LOG.info("Running expired registration request cleanup with expiry days= {}", expiryDays);
+    Instant expiryTime = Instant.now().minus(expiryDays, ChronoUnit.DAYS);
+    registrationRequestRepository.findExpiredRequests(Date.from(expiryTime))
+        .forEach(registrationRequestService::timeoutRequest);
```
enricovianello, line 166: 💯

enricovianello (changes requested):
Minor fixes to default values.
A part of the old implementation is still present but could be removed because it's uses only by tests.
Tests needs to be fixed properly as soon as the old code is removed.
Except for these things, the PR can be considered approved 👍

iam-login-service/src/main/java/it/infn/mw/iam/config/TaskConfig.java
garaimanoj: Done

iam-login-service/src/main/java/it/infn/mw/iam/config/TaskConfig.java
garaimanoj: Done

iam-login-service/src/main/java/it/infn/mw/iam/config/TaskConfig.java
garaimanoj: Done

iam-login-service/src/main/java/it/infn/mw/iam/core/user/DefaultIamAccountService.java
garaimanoj: Removed

iam-login-service/src/main/java/it/infn/mw/iam/core/user/IamAccountService.java
garaimanoj: Removed

garaimanoj:
> Minor fixes to default values. A part of the old implementation is still present but could be removed because it's uses only by tests. Tests needs to be fixed properly as soon as the old code is removed. Except for these things, the PR can be considered approved 👍

Ah I missed that one. Thanks for your review.

enricovianello (approved):

PR #1243, opened 2026-05-25
Search for the SAML attribute by alias as well

### LINKED ISSUE #1242: SAML JIT account provisioning fails if the assertion returns "friendly name" as the attribute name
SAML assertion returns the “friendly name” as the _attribute name_, whereas the code checks for the corresponding OID URI (e.g., `urn:oid:0.9.2342.19200300.100.1.3` => `mail`).

It could be related to https://github.com/indigo-iam/iam/issues/753.

rmiccoli:


iam-login-service/src/main/java/it/infn/mw/iam/authn/saml/JustInTimeProvisioningSAMLUserDetailsService.java
```diff
@@ -134,11 +131,15 @@ private void mapAttributes(SAMLCredential credential, IamSamlId samlId, IamAccou
...
 
-
   private String getAttributeAsString(SAMLCredential credential, Saml2Attribute attribute) {
-    return credential.getAttributeAsString(attribute.getAttributeName());
-  }
+    String value = credential.getAttributeAsString(attribute.getAttributeName());
+
+    if (value == null) {
+      value = credential.getAttributeAsString(attribute.getAlias());
+    }
```
enricovianello: I think the same logic should be put inside SamlExternalAuthenticationToken.toExernalAuthenticationRegistrationInfo() method in order to look for an alias match also in case of a registration request.

enricovianello (commented):
I think this could be the fix of a broken login match.
I suggest to check if we need to update also the registration stuff (e.g. SamlExternalAuthenticationToken.toExernalAuthenticationRegistrationInfo method implementation)

iam-login-service/src/main/java/it/infn/mw/iam/authn/saml/util/Saml2Attribute.java
```diff
@@ -73,12 +76,21 @@ public static Optional<Saml2Attribute> byName(String name) {
...
+    throw new IllegalArgumentException("Unknown SAML 2 attribute: " + alias);
+  }
+
+  public String resolveValue(SAMLCredential credential) {
+    String value = credential.getAttributeAsString(getAttributeName());
+
+    if (value == null) {
+      value = credential.getAttributeAsString(getAlias());
+    }
+    return value;
```
enricovianello: I vote for an Optional<String> for the returned value. Something similar to the **resolveAttribute** below.
I suggest something like:

```
Optional<String> value = Optional.ofNullable(credential.getAttributeAsString(getAttributeName()));
if (value.isEmpty()) {
  return Optional.ofNullable(credential.getAttributeAsString(getAlias()));
}
return value;
```

enricovianello (commented):
Just a minor fix. But I need to investigate a bit if it's the case to change the different behavior between Saml2Attribute.byName and Saml2Attribute.byAlias (could throw an exception)

enricovianello (approved):

PR #1251, opened 2026-06-10
Handle count zero for group member api call

### LINKED ISSUE #1204: SCIM group member resource returns a 500 error code when passing count=0 as a parameter
Using the API, to work with pagination it is sometimes useful to read just the number of resources. The quickest and more efficient way is to issue a get request specifying `0` as the `count` url parameter, like:

```console
$ curl -s -H "Autho..." https://iam-base-url/scim/Groups?count=0
{"totalResults":55,"schemas":["urn:ietf:params:scim:api:messages:2.0:ListResponse"],"Resources":[]}
```

While this works fine with the `scim/Groups` endpoint, it does not with the `scim/Groups/{id}/members` one. In the latter case, the server responses with a `500 Server Error`

To replicate:

```
curl -s -H "Autho..." https://iam-base-url/scim/Groups/{id}/members?count=0
```

Tested on 1.13.4.

garaimanoj:
No more 500 error with count=0

image

### DIFF
--- a/iam-login-service/src/test/java/it/infn/mw/iam/test/scim/group/ScimGroupManagerTests.java
+++ b/iam-login-service/src/test/java/it/infn/mw/iam/test/scim/group/ScimGroupManagerTests.java
@@ -130,6 +130,16 @@ class ScimGroupManagerTests {
   }
 
   @Test
+  @WithMockUser(username = "test", roles = "READER")
+  void roleReaderCanSeeGroupMembersCount() throws Exception {
+    mvc.perform(get(GROUP_URI + "/{uuid}/members?count=0", TEST_001_GROUP_ID).content(SCIM_CONTENT_TYPE))
+      .andExpect(status().isOk())
+      .andExpect(content().contentType(SCIM_CONTENT_TYPE))
+      .andExpect(jsonPath("$.totalResults", equalTo(0)))
+      .andExpect(jsonPath("$.Resources").isArray());
+  }
+
+  @Test
   @WithMockUser(username = "test", roles = {"USER", "GM:" + TEST_001_GROUP_ID})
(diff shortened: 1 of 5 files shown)


rmiccoli (approved):

iam-login-service/src/test/java/it/infn/mw/iam/test/scim/group/ScimGroupManagerTests.java
```diff
@@ -129,6 +129,16 @@ void roleReaderCanSeeGroupMembers() throws Exception {
...
 
+  @Test
+  @WithMockUser(username = "test", roles = "READER")
+  void roleReaderCanSeeGroupMembersCount() throws Exception {
+    mvc.perform(get(GROUP_URI + "/{uuid}/members?count=0", TEST_001_GROUP_ID).content(SCIM_CONTENT_TYPE))
+      .andExpect(status().isOk())
+      .andExpect(content().contentType(SCIM_CONTENT_TYPE))
+      .andExpect(jsonPath("$.totalResults", equalTo(0)))
+      .andExpect(jsonPath("$.Resources").isArray());
+  }
```
enricovianello, line 141: We could add another call specifying Analisys or Production group uuid because these groups have a non-empty amount of members

enricovianello (changes requested):
Just a copy and paste in the test, in order to add a test for a non empty group.

iam-login-service/src/test/java/it/infn/mw/iam/test/scim/group/ScimGroupManagerTests.java
garaimanoj: Done

enricovianello (approved):

PR #1252, opened 2026-06-11
Implement a simple callout to OPA

### LINKED ISSUE #1220: Implement a simple callout to OPA
This includes that OPA itself contains the static policies in a data.yml file, which is for now not updated.

We should also add some OPA property, e.g.

```bash
iam:
  opa:
    enable: false
    host: localhost
```

EKACRG:
If we set the property `IAM_OPA_ENABLED=true`, IAM will use OPA to evaluate the scope policy. In case of unsuccessful OPA response, an error is thrown.

A typical IAM input for OPA is something like

```bash
{
        "user": {
            "id": "73f16d93-2441-4a50-88ff-85360d78c6b5",
            "groups": [
                "31d9230c-90ae-4457-a990-0c443ab4aacc",
                "c617d586-54e6-411d-8e38-649677980004"
            ]
        },
        "client": {
             "id": "71e2fe91-1f85-4333-9098-406a1a36292a"
        },
        "scopes": [
            "openid",
            "storage.read:/pippo/pluto/paperino"
        ]
}
```

The IAM policies are exposed as OPA can understand them at the `/iam/scope_policies/opa` endpoint (requires Admin privileges).

In this PR we also allow filtering scopes based on the _clientId_. From OPA side, the policy evaluation order of a Client is the same as an account (i.e. it overrides group policies) . The _clientId_ is propagated also to the old engine, even if it is not evaluated.

We have added here an option to early fail in case some requested scope is not allowed -- not enabled by default.


iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/scope/pdp/OpaScopePolicyEngine.java
```diff
@@ -0,0 +1,83 @@
...
+    } else {
+      return super.apply(requestedScopes, account);
+    }
+  }
+
+  public Optional<OpaResponse> evaluatePolicy(@RequestBody OpaRequest payload) {
+    try {
+      String opaUrl = opaProperties.getUrl();
+      ResponseEntity<OpaResponse> response = restTemplate.postForEntity(opaUrl, payload, OpaResponse.class);
+      System.out.println(response);
```
enricovianello, line 71: As suggested by Sonar, the logger defined above is a better choice in this case.
I'd use the debug level. e.g.

```
LOG.debug("Response: {}", response);
```

enricovianello (approved):

PR #1271, opened 2026-06-23
Add grantType parameter for token events

### LINKED ISSUE #1226: A refresh token request logs a different grant_type
To reproduce it, I've obtained a RT with authorization code flow and then applying a refresh token flow IAM logs an `AccessTokenIssuedEvent` as follows

```bash
{"@type":"AccessTokenIssuedEvent","timestamp":1777465495770,"category":"TOKEN","principal":"client","message":"Issue access token","scopes":["iam:admin.read","iam:admin.write","openid","email","profile","offline_access"],"subject":"admin","grantType":"authorization_code","header":{"kid":"rsa1","alg":"RS256"},"payload":{"sub":"73f16d93-2441-4a50-88ff-85360d78c6b5","iss":"http://localhost:8080/","iat":1777465495732,"exp":1777469095732,"jti":"06137a7b-091a-4ffa-987d-274884f40481","client_id":"client","aud":["http://example1.com","http://example2.com","http://example3.com"],"scope":"openid profile offline_access iam:admin.read iam:admin.write email"},"refreshTokenJti":"f6f79822-8d04-4650-bb8e-d00667b7dc36","source":"IamTokenService"}
```

Basically, the grant_type results the one of the first authorization request.

We need to change the [TokenEvent](https://github.com/indigo-iam/iam/blob/develop/iam-login-service/src/main/java/it/infn/mw/iam/audit/events/tokens/TokenEvent.java#L50) class in order to retrieve the information from `getRefreshTokenRequest()` in the `OAuth2Request` object, if any.

garaimanoj:
After the change:

`{"@type":"AccessTokenIssuedEvent","timestamp":1782483755375,"category":"TOKEN","principal":"password-grant","message":"Issue access token","scopes":["storage.read:/","openid","offline_access"],"subject":"test","grantType":"refresh_token","header":{"kid":"rsa1","alg":"RS256"},"payload":{"sub":"80e5fb8d-b7c8-451a-89ba-346ae278a66f","iss":"http://localhost:8080/","iat":1782483738755,"exp":1782487338755,"jti":"d6182dbe-ecb1-4e91-9794-64c5579cc2dd","client_id":"password-grant"},"refreshTokenJti":"de8806ad-8c63-450c-99f2-a110c507e4d6","source":"IamTokenService"}
`

### DIFF
--- a/iam-login-service/src/main/java/it/infn/mw/iam/core/as/IamAuthorizationServerTokenServices.java
+++ b/iam-login-service/src/main/java/it/infn/mw/iam/core/as/IamAuthorizationServerTokenServices.java
@@ -423,6 +423,8 @@ public class IamAuthorizationServerTokenServices implements AuthorizationServerT
     
     OAuth2Authentication newOAuth2Authentication =
         new OAuth2Authentication(newOAuth2Request, authHolder.getUserAuth());
+    
+    AuthenticationHolderEntity newAuthHolder = createAuthenticationHolder(newOAuth2Authentication);    
 
     JWTProfile profile = profileResolver.resolveProfile(client.getScope());
 
@@ -476,7 +478,7 @@ public class IamAuthorizationServerTokenServices implements AuthorizationServerT
       revocationService.revokeRefreshToken(refreshToken);
     }
 
-    token.setAuthenticationHolder(authHolder);
+    token.setAuthenticationHolder(newAuthHolder);
 
 
     JWTClaimsSet atClaims = profile.getAccessTokenBuilder()
--- a/iam-login-service/src/main/java/it/infn/mw/iam/core/as/IamAuthorizationServerTokenServices.java
+++ b/iam-login-service/src/main/java/it/infn/mw/iam/core/as/IamAuthorizationServerTokenServices.java
@@ -204,7 +204,7 @@ public class IamAuthorizationServerTokenServices implements AuthorizationServerT
     if (client.isAllowRefresh()
         && isRefreshTokenRequested(request.getGrantType(), accessToken.getScope())) {
 
-      accessToken.setRefreshToken(createRefreshToken(client, authHolder));
+      accessToken.setRefreshToken(createRefreshToken(client, authHolder, request.getGrantType()));
     }
 
     JWTProfile profile = profileResolver.resolveProfile(client.getScope());
@@ -219,7 +219,7 @@ public class IamAuthorizationServerTokenServices implements AuthorizationServerT
 
       JWT idToken =
           createIdToken(client, request, Date.from(iat), account.get().getUuid(), accessToken);
-      eventPublisher.publishEvent(new IdTokenIssuedEvent(this, idToken, authHolder));
+      eventPublisher.publishEvent(new IdTokenIssuedEvent(this, idToken, authHolder, request.getGrantType()));
       accessToken.setIdToken(idToken);
     }
 
@@ -239,7 +239,7 @@ public class IamAuthorizationServerTokenServices implements AuthorizationServerT
   }
 
   private OAuth2RefreshTokenEntity createRefreshToken(ClientDetailsEntity client,
-      AuthenticationHolderEntity authHolder) {
+      AuthenticationHolderEntity authHolder, String grantType) {
 
     String jti = UUID.randomUUID().toString();
     Instant iat = clock.instant();
@@ -265,7 +265,7 @@ public class IamAuthorizationServerTokenServices implements AuthorizationServerT
     refreshToken.setClient(client);
 
     refreshToken = saveRefreshToken(refreshToken);
-    eventPublisher.publishEvent(new RefreshTokenIssuedEvent(this, refreshToken));
+    eventPublisher.publishEvent(new RefreshTokenIssuedEvent(this, refreshToken, grantType));
 
     return refreshToken;
   }
@@ -471,7 +471,7 @@ public class IamAuthorizationServerTokenServices implements AuthorizationServerT
       token.setRefreshToken(refreshToken);
     } else {
       // otherwise, make a new refresh token
-      token.setRefreshToken(createRefreshToken(client, authHolder));
+      token.setRefreshToken(createRefreshToken(client, authHolder, authRequest.getGrantType()));
       // clean up the old refresh token
       revocationService.revokeRefreshToken(refreshToken);
     }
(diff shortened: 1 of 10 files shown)


enricovianello (commented):
What about putting inside the TokenEvent a check:

```
if (authenticationHolder.getAuthentication().getOAuth2Request().isRefresh()) {
  this.grantType = AuthorizationGrantType.REFRESH_TOKEN.getGrantType();
} else {
  this.grantType = authenticationHolder.getAuthentication().getOAuth2Request().getGrantType();
}
```

We only need to change what we're logging in the end.

garaimanoj:
> What about putting inside the TokenEvent a check:
> 
> ```
> if (authenticationHolder.getAuthentication().getOAuth2Request().isRefresh()) {
>   this.grantType = AuthorizationGrantType.REFRESH_TOKEN.getGrantType();
> } else {
>   this.grantType = authenticationHolder.getAuthentication().getOAuth2Request().getGrantType();
> }
> ```
> 
> We only need to change what we're logging in the end.

Thanks @enricovianello , I have tried that solution but the `refresh` is null in `OAuth2Request`.

image

enricovianello:
> > What about putting inside the TokenEvent a check:
> > ```
> > if (authenticationHolder.getAuthentication().getOAuth2Request().isRefresh()) {
> >   this.grantType = AuthorizationGrantType.REFRESH_TOKEN.getGrantType();
> > } else {
> >   this.grantType = authenticationHolder.getAuthentication().getOAuth2Request().getGrantType();
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We only need to change what we're logging in the end.
> 
> Thanks @enricovianello , I have tried that solution but the `refresh` is null in `OAuth2Request`.
> 
> image

That's true. The authenticationHolder holds the original request. And that's why I think that authenticationHolder should stay what it is. At this point the only approach I think is to add to TokenEvent (and also Access|RefreshTokenEvent) a grantType parameter. Inside IamTokenService there are 3 points where this events are called and we can get the grantType from the current OAuth2Request or from TokenRequest. Inside "refreshAccessToken" method I think we have no doubt the grantType will be the refresh_token one.

What do you think?

garaimanoj:
> > > What about putting inside the TokenEvent a check:
> > > ```
> > > if (authenticationHolder.getAuthentication().getOAuth2Request().isRefresh()) {
> > >   this.grantType = AuthorizationGrantType.REFRESH_TOKEN.getGrantType();
> > > } else {
> > >   this.grantType = authenticationHolder.getAuthentication().getOAuth2Request().getGrantType();
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > We only need to change what we're logging in the end.
> > 
> > 
> > Thanks @enricovianello , I have tried that solution but the `refresh` is null in `OAuth2Request`.
> > image
> 
> That's true. The authenticationHolder holds the original request. And that's why I think that authenticationHolder should stay what it is. At this point the only approach I think is to add to TokenEvent (and also Access|RefreshTokenEvent) a grantType parameter. Inside IamTokenService there are 3 points where this events are called and we can get the grantType from the current OAuth2Request or from TokenRequest. Inside "refreshAccessToken" method I think we have no doubt the grantType will be the refresh_token one.
> 
> What do you think?

I’ve already taken similar approach and have started running the test. Thanks for your idea though.

iam-login-service/src/main/java/it/infn/mw/iam/core/IamTokenService.java
enricovianello: During refresh flow, when a new refresh token is created, I think we'll use the original refreshToken AuthenticationHolder to get the grantType. And maybe this won't fix the problem. And we cannot solve this problem because that createRefreshToken method has missing parameters.
That method comes from mitreId interface:
https://github.com/indigo-iam/OpenID-Connect-Java-Spring-Server/blob/v2/openid-connect-common/src/main/java/org/mitre/oauth2/service/OAuth2TokenEntityService.java

Maybe some tests will highlight my fear.

But..
I'm decoupling the codebase from this interface in another PR https://github.com/indigo-iam/iam/pull/1274.
It's a big refactoring. In practice IamTokenService has been splitted into 2 objects to implement the 2 interfaces of the framework:
- `AuthorizationServerTokenServices` https://github.com/indigo-iam/iam/pull/1274/changes#diff-4d8c0d6bbfbcfc45a6d6d553d51f6249431ac692d1ace2afe9666b979c0f6c8f
- `ResourceServerTokenServices` https://github.com/indigo-iam/iam/pull/1274/changes#diff-ebd46480907ae8440096c0dfa04df227f37c28c31a1e5623b58e928803cfcb11

So, probably, if I merge my PR before and you update this branch, the fix becomes easier because we'll have the freedom to review the flow without being strict to the MitreId interface.

Another strategy:
- you can add the tests here and verify if I'm right or wrong
- if I'm right, and tests are failing, I could merge your PR inside mine e fix those 2-3 necessary lines!

I don't know if there are simpler approach in that case.
If I'm wrong, I'll merge this asap :)

iam-login-service/src/main/java/it/infn/mw/iam/core/as/IamAuthorizationServerTokenServices.java
```diff
@@ -228,7 +228,7 @@ && isRefreshTokenRequested(request.getGrantType(), accessToken.getScope())) {
     }
 
     OAuth2AccessTokenEntity savedAccessToken = saveAccessToken(accessToken);
-    eventPublisher.publishEvent(new AccessTokenIssuedEvent(this, savedAccessToken));
+    eventPublisher.publishEvent(new AccessTokenIssuedEvent(this, savedAccessToken, authentication.getOAuth2Request().getGrantType()));
```
enricovianello: `request.getGrantType()` as above

enricovianello (changes requested):
It's approved after this very very minor fix 👍

iam-login-service/src/main/java/it/infn/mw/iam/core/as/IamAuthorizationServerTokenServices.java
garaimanoj: Done

enricovianello (approved):

PR #1277, opened 2026-06-24
Update voperson_id and eduPersonScopedAffiliation to use config value or the issuer

### LINKED ISSUE #1276: Clarify scoped identifier formats for voperson_id and eduPersonScopedAffiliation
Current documentation/examples for scoped identifiers mix different patterns without clearly distinguishing their semantics. In particular:

- The `voperson_id` example uses a value like:

  ```json
  "voperson_id": "sub@iam_organization_name"
  ```

- The `eduPersonScopedAffiliation` example uses a value like:
     ```json
    "eduPersonScopedAffiliation": "affiliation@iam_organization_name"  
    ```
In both cases, the right‑hand side (`iam_organization_name`) is taken from an organisation name that may contain spaces and is not guaranteed to be a globally unique, domain‑style scope.

### Problem
-  The current `iam_organization_name` value can contain spaces and other characters that are problematic for scoped identifiers and may not conform to existing eduPerson / voPerson guidance for scope values
- Because the scope part is not a stable, globally unique identifier (such as a domain name or issuer URL), different organisations or deployments may end up using overlapping or ambiguous scopes. This breaks the intent of scoped identifiers, where the right‑hand side should clearly identify the authority or domain that issued the identifier. 

### Proposed solutions
1. **Introduce a separate configurable scope value**

   - Add configuration options for the scope/authority part used after `@` in both `voperson_id` and `eduPersonScopedAffiliation`.
   - This scope value should be something that is globally unique.

   For example:
  
   ```json
   "voperson_id": "sub@https://iam.example.org"
   "eduPersonScopedAffiliation": "member@iam.example.org"
   ```
  
   where `https://iam.example.org` (or `iam.example.org`) is explicitly configured rather than derived from a display name.

2. **Change the default scope value**

   - If no explicit configuration is provided, change the default right‑hand side to something like the OIDC issuer (`iss`) or a related domain value, instead of the human‑readable organisation name

anhau:
Currently indigio iam is using the `iam_organisation_name` as the affiliation scope for voperson_id and eduPersonScopedAffiliation. This pull request introduces a new feature to set:
  ```json
aarc-profile:
    affiliation-scope: ${IAM_AARC_PROFILE_AFFILIATION_SCOPE:iam_example}
```
Which will change the behavior of  `voperson_id` to be `sub@affiliation_scope` and `eduPersonScopedAffiliation`  to be `affiliation@affiliation_scope`

If `affiliation_scope` is not set and empty, the issuer will instead be used on the right side of the "@", such as sub@issuer and affiliation@issuer to make sure these values are globally unique

### DIFF
--- a/iam-login-service/src/test/java/it/infn/mw/iam/test/oauth/profile/AarcClaimValueHelperTests.java
+++ b/iam-login-service/src/test/java/it/infn/mw/iam/test/oauth/profile/AarcClaimValueHelperTests.java
@@ -20,7 +20,6 @@ import static org.hamcrest.CoreMatchers.hasItem;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.empty;
 import static org.hamcrest.Matchers.hasSize;
-import static org.hamcrest.Matchers.equalTo;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
@@ -168,8 +167,8 @@ class AarcClaimValueHelperTests {
     String scopedAffiliation = (String) helper.resolveClaim(
         AarcExtraClaimNames.EDUPERSON_SCOPED_AFFILIATION, null, Optional.of(account));
 
-    assertThat(vopersonId, equalTo("test-subject@https://scope.example"));
-    assertThat(scopedAffiliation, equalTo("member@https://scope.example"));
+    assertEquals(vopersonId, "test-subject@https://scope.example");
+    assertEquals(scopedAffiliation, "member@https://scope.example");
   }
 
   @Test
@@ -188,8 +187,8 @@ class AarcClaimValueHelperTests {
     String scopedAffiliation = (String) helper.resolveClaim(
         AarcExtraClaimNames.EDUPERSON_SCOPED_AFFILIATION, null, Optional.of(account));
 
-    assertThat(vopersonId, equalTo("test-subject@https://issuer.example"));
-    assertThat(scopedAffiliation, equalTo("member@https://issuer.example"));
+    assertEquals(vopersonId, "test-subject@https://issuer.example");
+    assertEquals(scopedAffiliation, "member@https://issuer.example");
   }
 
   @Test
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: iam-login-service/src/test/java/it/infn/mw/iam/test/oauth/profile/AarcClaimValueHelperTests.java [hunk 1]
```diff
@@ -151,6 +152,46 @@ void testEmptyGroupListEncode() {
...
+
+    when(account.getUserInfo()).thenReturn(accountUserInfo);
+    when(accountUserInfo.getSub()).thenReturn("test-subject");
+
+    String vopersonId = (String) helper.resolveClaim(AarcExtraClaimNames.VOPERSON_ID, null,
+        Optional.of(account));
+    String scopedAffiliation = (String) helper.resolveClaim(
+        AarcExtraClaimNames.EDUPERSON_SCOPED_AFFILIATION, null, Optional.of(account));
+
+    assertThat(vopersonId, equalTo("test-subject@https://scope.example"));
```
rmiccoli: You can use the already imported `assertEquals`

iam-login-service/src/test/java/it/infn/mw/iam/test/oauth/profile/AarcClaimValueHelperTests.java
```diff
@@ -151,6 +152,46 @@ void testEmptyGroupListEncode() {
...
+
+    when(account.getUserInfo()).thenReturn(accountUserInfo);
+    when(accountUserInfo.getSub()).thenReturn("test-subject");
+
+    String vopersonId = (String) helper.resolveClaim(AarcExtraClaimNames.VOPERSON_ID, null,
+        Optional.of(account));
+    String scopedAffiliation = (String) helper.resolveClaim(
+        AarcExtraClaimNames.EDUPERSON_SCOPED_AFFILIATION, null, Optional.of(account));
+
+    assertThat(vopersonId, equalTo("test-subject@https://issuer.example"));
```
rmiccoli: Same as above

rmiccoli (changes requested):
LGTM 👍 
Just a very minor fix

rmiccoli (approved):

enricovianello (approved):

PR #1293, opened 2026-07-03
Add iam-voms-aa RPM (and refactor RPM build)

### LINKED ISSUE #1291: Feature request: rpms for iam-voms-aa
We need to roll out the voms-aa service on our rpm based IAM server, hence the need for an rpm for this service.
Note: Simon F (@sfayer) will provide the rpms, if possible assign it to him directly.

sfayer:
Hi,

Here is our proposal for adding the iam-voms-aa RPM: I've done quite a large amount of refactoring of the other RPM packages so that a single spec file now builds an indigo-iam-login-service, indigo-iam-voms-aa and a top level package that owns the container directory and "iam" user (indigo-iam). I've also moved some of the files around and changed the permissions to match up with the standard Redhat packaging guidelines. For the unit files, I've added some "generic" ones that can take a paramter: this allows multiple instances of these services to be run on a single host (`systemctl start iam-voms-aa@myvo.service`) just like the VOMS packages used to do.

Finally, I've updated the workflow script so that it uploads the full set of RPMs rather than just the one: I hope that's OK, I can't test that bit: I can remove it again if needed.

Please let me and @marianne013 know what you think!

Closes #1291.

Regards,
Simon

### DIFF
--- a/.github/workflows/build-rpm.yaml
+++ b/.github/workflows/build-rpm.yaml
@@ -108,4 +108,9 @@ jobs:
       - name: Upload release to repo
         if: env.REPO != ''
         run: |
-          curl --fail --user "${{ vars.NEXUS_USERNAME }}:${{ secrets.NEXUS_PASSWORD }}" --upload-file rpmbuild/RPMS/noarch/*.rpm https://repo.cloud.cnaf.infn.it/repository/indigo-iam-rpm-${{ env.REPO }}/almalinux${{ matrix.version }}/
+          for f in rpmbuild/RPMS/noarch/*.rpm; do
+            curl --fail \
+              --user "${{ vars.NEXUS_USERNAME }}:${{ secrets.NEXUS_PASSWORD }}" \
+              --upload-file "$f" \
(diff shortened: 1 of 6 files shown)


expuss2000:
Hi, thank you for your contribution.
It all seems reasonable to me, the only minor change that I would make to follow the [Java Packaging Guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/Java/#_jar_file_installation) would be to use the `%{_javadir}` (so `/usr/share/java`) instead of `%{_datadir}`.
Also I'm not sure whether the JAR file in the old location will be deleted.

sfayer:
Hi,

Thanks. I did consider using `%{__javadir}` but decided that these would be best placed under a datadir following the "application-specific JAR files" rule (as the jar/war file are the full application rather than library files likely to be referenced by other applications). I can of-course change this if you still think javadir would be better?

Regards,
Simon

expuss2000:
Hi, that makes sense. I don't know how much this exception is followed because in my current installation (so it could be a coincidence) all JAR files are under `/usr/share/java` (except for Maven libraries), but I found, for example, that Tomcat follows it.

enricovianello (approved):

PR #1306, opened 2026-07-10
URL-encode/decode client credentials in Authorization header

### LINKED ISSUE #1305: Support OAuth2.1 URL encoding/decoding of client credentials
Ref: https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1#section-2.4.1

rmiccoli:
Following [OAuth2.1](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1#section-2.4.1)


enricovianello (changes requested):
We could add a client with a url-encoded client-id in our test data and then add a test of a call to the token endpoint with basic auth 👍

enricovianello (approved):

PR #1316, opened 2026-07-24
Hash client secret with HMAC-SHA256

federicaagostini:
This PR superseeds https://github.com/indigo-iam/iam/pull/947.

Right now, it is missing the db migration for secrets already saved in the db.

Evaluate also the usage of SHA256 alone.


giacomini:
I don't see any compelling reason to use hmac or even encryption over a simple cryptographic hash like sha256 or sha512. Searching for other experiences I've only found references to hashes:
- https://docs.gitlab.com/integration/oauth_provider/#hashed-oauth-application-secrets
- https://support.google.com/cloud/answer/15549257?hl=en (note here that they are able to display the last four characters of the client secret, just for secret identification purposes, which seems a nice feature for the users)
- https://groups.google.com/g/keycloak-dev/c/TmsNfnol0_g?pli=1
- https://www.oauth.com/oauth2-servers/client-registration/client-id-secret/

PR #1318, opened 2026-07-27
Hash client secrets and add benchmark for encoder

### LINKED ISSUE #923: Client secret hashed on DataBase
The client secret is currently stored on database without any encryption, then it should be hashed for security reasons.
This means that the API must ignore any secret received during the creation/update of a client, generate it on POST creation requests and provide it to the user/dashboard one time in the relative response.
If the client's owner loses his client secret, the dashboard will provide him a button to  invalidate the current one and generate a new one (returned in the response).
Within the database, the client secret column type won't change but it will be stored as a bcrypt hash instead.
A proper migration that encrypts current client secrets is necessary.

federicaagostini:
In this PR we hash the client secrets with SHA256, superseeding PR https://github.com/indigo-iam/iam/pull/1316.

We have tested a db migration with 58935 clients and it takes around 14 s (on my laptop, where the db is deployed with docker-compose), which could be feasible. The following query compares this migration with the slowest ones:

`SELECT execution_time,script,installed_rank,version,type,description FROM schema_version ORDER BY execution_time DESC LIMIT 10;`
| execution_time | script | installed_rank | version | type | description |
|---:|---|---:|---:|---|---|
| 87779 | `V96__add_foreign_keys.sql` | 53 | 96 | SQL | add foreign keys |
| 49207 | `V96__add_foreign_keys.sql` | 95 | 96 | SQL | add foreign keys |
| 22011 | `db.migration.mysql.V81_2___RemoveOrphanTokens` | 46 | 81.2 | JDBC | RemoveOrphanTokens |
| 19506 | `db.migration.mysql.V81_2___RemoveOrphanTokens` | 88 | 81.2 | JDBC | RemoveOrphanTokens |
| 18560 | `V93__add_at_hash.sql` | 50 | 93 | SQL | add at hash |
| 17235 | `V100__alter_saved_user_auth_info_table.sql` | 57 | 100 | SQL | alter saved user auth info table |
| 13727 | `db.migration.mysql.V123__HashClientSecret` | 122 | 123 | JDBC | HashClientSecret |
| 13551 | `V1___init.sql` | 1 | 1 | SQL | init |
| 12242 | `V2___iam_tables.sql` | 2 | 2 | SQL | iam tables |
| 12221 | `V93__add_at_hash.sql` | 92 | 93 | SQL | add at hash |

In this PR we also calculate with the Java Microbenchmark Harness how much different alogrithms take to check if an encoded secret matches with a plain secret (which is the use case of basic authentication for instance) and we get an output like

```
Benchmark                       (algorithm)  Mode  Cnt      Score      Error  Units
PasswordEncoderBenchmark.match       BCRYPT  avgt   20  66217.124 ± 1680.047  us/op
PasswordEncoderBenchmark.match       PBKDF2  avgt   20  76889.045 ± 3492.578  us/op
PasswordEncoderBenchmark.match         HMAC  avgt   20      0.677 ±    0.033  us/op
PasswordEncoderBenchmark.match       SHA256  avgt   20      0.246 ±    0.016  us/op
```


iam-login-service/src/test/java/it/infn/mw/iam/test/oauth/assertion/JWTBearerClientAuthenticationIntegrationTestSupport.java
```diff
@@ -43,7 +44,8 @@
 public class JWTBearerClientAuthenticationIntegrationTestSupport extends EndpointsTestUtils {
 
   public static final String CLIENT_ID_SECRET_JWT = "jwt-auth-client_secret_jwt";
-  public static final String CLIENT_ID_SECRET_JWT_SECRET = "c8e9eed0-e6e4-4a66-b16e-6f37096356a7";
+  public static final String CLIENT_ID_SECRET_JWT_SECRET =
+      new IamSha256PasswordEncoder().encode("c8e9eed0-e6e4-4a66-b16e-6f37096356a7");
```
enricovianello: The test that uses this encoded secret loses its sense:

```
@Test
  void testSymmetricJwtAuthHasBeenDeprecated() throws Exception {

    JWT jwt = createSymmetricClientAuthToken(CLIENT_ID_SECRET_JWT, Instant.now().plusSeconds(600));
    String serializedToken = jwt.serialize();

    mvc
      .perform(post(TOKEN_ENDPOINT).param("client_id", CLIENT_ID_SECRET_JWT)
        .param("client_assertion_type", JWT_BEARER_ASSERTION_TYPE)
        .param("client_assertion", serializedToken)
        .param("grant_type", "client_credentials"))
      .andExpect(status().isUnauthorized())
      .andExpect(jsonPath("$.error_description", is("SECRET_JWT has been deprecated")));
  }
```

Because you're now authN with the hashed secret. You need 2 variables: one for the plain secret and another with the encoded one.

federicaagostini: I went back with the plain secret since the error "SECRET_JWT has been deprecated" is shown anyhow.

PR #1322, opened 2026-08-12
Fix token exchange requested token type handling

enricovianello:
## Motivation

The INDIGO IAM token exchange implementation currently does not take the `requested_token_type` parameter into account when processing an OAuth 2.0 Token Exchange request.

Regardless of the requested token type, IAM currently issues an OAuth access token and reports:

```json
{
  "access_token": "<access-token>",
  "token_type": "Bearer",
  "issued_token_type": "urn:ietf:params:oauth:token-type:jwt"
}
```

A refresh token may additionally be returned when the requested scopes and client configuration allow one.

This behavior is problematic when a client explicitly requests a different token type, in particular:

```text
requested_token_type=urn:ietf:params:oauth:token-type:refresh_token
```

According to [RFC 8693](https://datatracker.ietf.org/doc/html/rfc8693), `requested_token_type` indicates the type of token the client wants the authorization server to issue. The token resulting from the exchange is returned in the `access_token` response parameter, even when the issued token is not actually an OAuth access token.

This somewhat counter-intuitive use of the `access_token` response parameter is explicitly defined by [RFC 8693](https://datatracker.ietf.org/doc/html/rfc8693).

## Changes

This change updates the token exchange implementation to honor `requested_token_type`.

The following token types are supported:

* `urn:ietf:params:oauth:token-type:access_token`
* `urn:ietf:params:oauth:token-type:refresh_token`

When `requested_token_type` is omitted, IAM defaults to issuing an OAuth access token, as specified by RFC 8693.

Unsupported requested token types are rejected.

The value of `issued_token_type` is also changed to describe the semantic type of the issued token rather than its JWT representation.

### Access token request

For a request such as:

```text
grant_type=urn:ietf:params:oauth:grant-type:token-exchange
subject_token=<subject-token>
subject_token_type=urn:ietf:params:oauth:token-type:jwt
requested_token_type=urn:ietf:params:oauth:token-type:access_token
```

IAM returns an OAuth access token:

```json
{
  "access_token": "<access-token>",
  "token_type": "Bearer",
  "issued_token_type": "urn:ietf:params:oauth:token-type:access_token",
  "expires_in": 3600
}
```

The same behavior applies when `requested_token_type` is omitted.

Previously, IAM reported:

```text
urn:ietf:params:oauth:token-type:jwt
```

as the `issued_token_type`.

The new value:

```text
urn:ietf:params:oauth:token-type:access_token
```

reflects that the issued token is being used as an OAuth access token, regardless of the fact that its concrete representation in IAM is a JWT.

### Refresh token request

For a request explicitly asking for a refresh token:

```text
grant_type=urn:ietf:params:oauth:grant-type:token-exchange
subject_token=<subject-token>
subject_token_type=urn:ietf:params:oauth:token-type:jwt
requested_token_type=urn:ietf:params:oauth:token-type:refresh_token
scope=openid offline_access
```

the refresh token is returned as the value of the `access_token` response parameter:

```json
{
  "access_token": "<refresh-token>",
  "token_type": "N_A",
  "issued_token_type": "urn:ietf:params:oauth:token-type:refresh_token",
  "expires_in": 2592000
}
```

`token_type` is set to `N_A` because the value carried by the `access_token` response parameter is not an OAuth access token and therefore cannot be used as a Bearer token.

The returned token can subsequently be used with the regular refresh token grant:

```text
grant_type=refresh_token
refresh_token=<value-returned-in-access_token>
```

to obtain a usable OAuth access token.

If IAM cannot issue a refresh token for the requested exchange, the request fails instead of silently returning an access token of a different type.

## Interoperability

This change also addresses an interoperability difference observed between INDIGO IAM and CILogon.

For a token exchange request containing:

```text
requested_token_type=urn:ietf:params:oauth:token-type:refresh_token
```

CILogon returns the requested refresh token in the `access_token` response parameter with:

```json
{
  "token_type": "N_A",
  "issued_token_type": "urn:ietf:params:oauth:token-type:refresh_token"
}
```

while INDIGO IAM currently returns a directly usable Bearer access token and, when applicable, a separate refresh token.

Clients consuming token exchange responses must therefore not assume that the `access_token` response parameter always contains a directly usable OAuth access token. They should take `issued_token_type` and `token_type` into account when processing the response.

## Backward compatibility

This introduces a behavioral change for token exchange clients.

Requests that omit `requested_token_type`, or explicitly request:

```text
urn:ietf:params:oauth:token-type:access_token
```

continue to receive a usable OAuth access token.

However, the `issued_token_type` for these responses changes from:

```text
urn:ietf:params:oauth:token-type:jwt
```

to:

```text
urn:ietf:params:oauth:token-type:access_token
```

More importantly, clients explicitly requesting:

```text
urn:ietf:params:oauth:token-type:refresh_token
```

will no longer receive a usable Bearer access token in the `access_token` response parameter. The field will instead contain the requested refresh token and `token_type` will be `N_A`.

Clients that currently request a refresh token but nevertheless consume `access_token` directly as a Bearer token will therefore need to be updated.

## Tests

The token exchange integration tests have been updated and extended to cover:

* default issuance of an access token when `requested_token_type` is omitted;
* explicit requests for an access token;
* explicit requests for a refresh token;
* correct `issued_token_type` values;
* `token_type=Bearer` for OAuth access tokens;
* `token_type=N_A` for issued refresh tokens;
* use of the refresh token returned in the `access_token` response parameter with the refresh token grant;
* rejection of unsupported requested token types;
* failure when a refresh token is requested but cannot be issued;
* existing audience, scope, impersonation and `act` claim behavior.

## References

[RFC 8693](https://datatracker.ietf.org/doc/html/rfc8693) — OAuth 2.0 Token Exchange, in particular sections 2.1 and 2.2.1.

### DIFF
--- a/iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/granters/TokenExchangeTokenGranter.java
+++ b/iam-login-service/src/main/java/it/infn/mw/iam/core/oauth/granters/TokenExchangeTokenGranter.java
@@ -19,12 +19,17 @@ import static it.infn.mw.iam.core.oauth.exchange.TokenExchangePdpResult.Decision
 import static it.infn.mw.iam.core.oauth.exchange.TokenExchangePdpResult.Decision.PERMIT;
 import static java.lang.String.format;
 
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Set;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.security.oauth2.client.resource.OAuth2AccessDeniedException;
+import org.springframework.security.oauth2.common.DefaultOAuth2AccessToken;
+import org.springframework.security.oauth2.common.ExpiringOAuth2RefreshToken;
 import org.springframework.security.oauth2.common.OAuth2AccessToken;
+import org.springframework.security.oauth2.common.OAuth2RefreshToken;
 import org.springframework.security.oauth2.common.exceptions.InvalidGrantException;
 import org.springframework.security.oauth2.common.exceptions.InvalidRequestException;
 import org.springframework.security.oauth2.common.exceptions.InvalidScopeException;
@@ -52,10 +57,18 @@ public class TokenExchangeTokenGranter extends AbstractTokenGranter {
 
   public static final Logger LOG = LoggerFactory.getLogger(TokenExchangeTokenGranter.class);
 
+  private static final String REQUESTED_TOKEN_TYPE = "requested_token_type";
+  private static final String ISSUED_TOKEN_TYPE = "issued_token_type";
+
+  private static final String ACCESS_TOKEN_TYPE =
+      "urn:ietf:params:oauth:token-type:access_token";
+  private static final String REFRESH_TOKEN_TYPE =
+      "urn:ietf:params:oauth:token-type:refresh_token";
+
+  private static final String TOKEN_TYPE_NA = "N_A";
+
   public static final String TOKEN_EXCHANGE_GRANT_TYPE =
       "urn:ietf:params:oauth:grant-type:token-exchange";
-
-  private static final String TOKEN_TYPE = "urn:ietf:params:oauth:token-type:jwt";
 
   private static final String AUDIENCE_FIELD = "audience";
   private static final String OFFLINE_ACCESS_SCOPE = "offline_access";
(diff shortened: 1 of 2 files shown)


maarten-litmaath:
Hi all,
will this change be compatible with what FTS-3 and -4 currently are doing in token exchange workflows?

If not, we would be unable to deploy the new code in production!

FTS-4 will be patched to do things right, while no code change is planned for FTS-3 anymore,
as it is expected to be phased out in the course of 2027.

Issue #1338, opened 2026-09-02
Prevent ordinary owners from enabling sensitive grants such as client_credentials


DonaldChung-HK:
I think this is already implemented for anonymous client but CTAO thinks that this should be a configurable option so that admin can make this grant type only available for admin user instead of ordinary users, or a list of grant type available for ordinary user to set up similar to how refresh token grant type is barred for ordinary users?

Thanks

PR #1359, opened 2026-09-17
Read forbidden grant types from configuration

basbs:
Replace the hardcoded enum with values read from
application.yaml under grant-types.forbidden-for-user-clients and grant-types.forbidden-for-anonymous-clients. This allows the forbidden grant type lists to be configured per environment instead of requiring a code change.

Refs #1338


basbs:
Hi @enricovianello, 

A couple of things I want to take your opinion on before making this PR ready for review, while we're introducing configurable forbidden grant types (replacing the previous hardcoded enum):

1. Pre-existing clients: clients created before this validation existed may already have grant types that are now forbidden, whether active or disabled. Not addressed in this PR, but something we should tackle.
2. Config drift: I kept the same values that previously existed in the enum, but since this is now driven by `application.yaml`, it's only as safe as whoever edits that file at deploy time. A typo or an unintentional override could silently weaken the forbidden list (I might do this myself for my own instance without realizing the impact). Worth discussing whether we want any safeguards here (e.g., startup validation, a non-overridable baseline, etc.) 

Let me know what you think.

